PR #177, opened 2014-03-15
Require Admin privileges while saving navigation bar configuration

### LINKED ISSUE #37: Navigation bar not showing
After i installed the extension, and restarted i could see the navigation bar.
After another restart of IDE, navigation bar disappeared forever...
Any ideas?

dungpa:
Fix #37.

@jaredpar Is there any way to turn on UAC or escalate permissions inside VS process? If yes, we might have better UI interaction than displaying an error message.

### DIFF
--- a/src/FSharpVSPowerTools/GeneralOptionsPage.cs
+++ b/src/FSharpVSPowerTools/GeneralOptionsPage.cs
@@ -21,10 +21,18 @@ namespace FSharpVSPowerTools
         
         private bool GetNavigationBarConfig()
         {
-            var b = ConfigurationManager.AppSettings.Get(navBarConfig);
-            bool result;
-            if (b != null && bool.TryParse(b, out result)) return result;
-            return false;
+            try
+            {
+                var config = ConfigurationManager.OpenExeConfiguration(Application.ExecutablePath);
+                var b = config.AppSettings.Settings[navBarConfig].Value;
+                bool result;
+                if (b != null && bool.TryParse(b, out result)) return result;
+                return false;
+            }
+            catch (Exception)
+            {
+                return false;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


xperiandri:
What does mean that Navigation Bar requires admin privileges?
Do I have to run VS with Admin Privileges once I want to enable it and that is enough?
Is there any explanation somewhere?

dungpa:
> Do I have to run VS with Admin Privileges once I want to enable it and that is enough?

Yes, that's enough.

> Is there any explanation somewhere?

http://tomasp.net/blog/regions-navigation.aspx/

xperiandri:
So why not to disable that option until Visual Studio is running with admin privileges?
I mean set enabled to combobox to false

dungpa:
Yes, that's better. Care to send a PR?

The whole logic is handled at https://github.com/fsprojects/VisualFSharpPowerTools/blob/b9373f46eddc2f5bfffde86d0df0748c7ef7fd27/src/FSharpVSPowerTools/GeneralOptionsPage.cs#L63-L78.

xperiandri:
Yes, I will

xperiandri:
How to make this if page will look like property page?
I mean that change that we want to implement.

dungpa:
What do you mean?

We replace `GeneralOptionsControl` by a wholly new control. `GeneralOptionsPage` should not change at all.
As long as you keep all public properties https://github.com/fsprojects/VisualFSharpPowerTools/blob/b9373f46eddc2f5bfffde86d0df0748c7ef7fd27/src/FSharpVSPowerTools/GeneralOptionsControl.cs#L15-L138 intact, the original behaviour is preserved.

xperiandri:
I mean how to disable an item in property grid for NavBar if not admin

PR #579, opened 2014-07-11
WIP: Rename suggestion list for rename dialog

forki:
This is WIP but I want to collect feedback from the beginning


vasily-kirichenko:
We should finally do something with it :)

dungpa:
The PR has been here for too long. I don't think we have resources to revive it. Closing.

PR #914, opened 2015-01-03
New FSharp.Formatting uses latest FCS

forki:
so they version conflict should be solved.


vasily-kirichenko:
Please remove https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/lib/FSharp.Compiler.Service.dll 

and fix `#I` here https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/docs/tools/generate.fsx#L28

forki:
done

vasily-kirichenko:
Great, thanks!

PR #902, opened 2014-12-21
Comment outlining

vasily-kirichenko:
This PR replaces https://github.com/fsprojects/VisualFSharpPowerTools/pull/875 as the author has abandoned it.

PR #919, opened 2015-01-13
Life easier for polyglot / C# / ES6 -developer to code F#

Thorium:
Some easy common typos are replaced while you write:
var to let,
C#/ES6 lambda to F#-lambda

### DIFF
--- a/docs/content/commoncstypos.fsx
+++ b/docs/content/commoncstypos.fsx
@@ -1,0 +1,17 @@
+(*** hide ***)
+// This block of code is omitted in the generated HTML documentation. Use 
+// it to define helpers that you do not want to show in the documentation.
+#I "../../bin"
+
+(**
+
+Auto-correct common C#-/JavaScript-developer typos in F#-files
+-----------------------
+
+The idea is to correct most common polyglot- / C#- / JavaScript-developer typing errors while they write F#:
+
+ - Writing only ``"var "`` in the beginning of the line, it will be replaced to ``"let "``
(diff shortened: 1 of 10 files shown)


Thorium:
Just try it. No need to break your flow when you accidentally write
```
var x = 3
```
to new empty line. Or if you accidentally forgot to write F#-style lambda after coding a lot of C# or ES6.

Thorium:
Hmm. I wonder if this is too naïve implementation for the lambda: F#-compiler-service is not used. So there could be cases that have custom operator (a <=> b) or you could have C#/JS-code in some kind of multi-line-string to convert it to something and then this could be just a little annoyance. You can enable/disable this from the project menu, but should the default be on or off... I think there are more newbies and non-primary-F# programmers than meta programmers.

vasily-kirichenko:
I think it should not be a part of VFPT since:

1. It's a very general feature. The same results one can easily achieve with an external system-wide tools like http://www.autohotkey.com/. 
2. It's not needed after you've been writing F# for at least a week.

Sorry, we have to reject this PR. If you want to help the project, you may take a look at https://github.com/fsprojects/VisualFSharpPowerTools/pull/902 which have severe performance issues and is not feature complete. 

For the future, please, discuss features/fixes in advance with @dungpa and me. Thanks.

Thorium:
Ok, no problem! :-)

dungpa:
@Thorium Sorry that it didn't work out for the first time. You're welcome to send bug fixes or implement feature requests from http://vfpt.uservoice.com/forums/247560-general/filters/top. However, due to nature of tooling development, it requires some discussion before proceeding.

PR #921, opened 2015-01-14
FsiReferenceCommand keep user defined references ordering

vasily-kirichenko:
update paket.bootstrapper.exe


src/FSharpVSPowerTools.Logic/FsiReferenceCommand.fs
```diff
@@ -141,29 +136,56 @@ type FsiReferenceCommand(dte2: DTE2, mcs: OleMenuCommandService, _shell: IVsUISh
...
+            let refs = 
+                match tryGetExistingFileRefs project loadRefsFileName with
+                | None -> actualRefs
+                | Some existingRefs ->
+                    // remove refs which don't actual anymore (they have been removed from the project)
+                    let existingRefs =
+                        existingRefs 
+                        |> List.filter (fun existingRef -> List.exists ((=) existingRef) actualRefs)
+                    // get refs which don't exist in the existing file
+                    let newExtraRefs =
```
dungpa: `newExtraRefs` is always empty since `existingRefs` is shadowed and contains only elements from `actualRefs`. So the code doesn't work if there is some added references that don't appear in newly generated script.

Probably remove (evil) value shadowing and calculate `survivedRefs` in the first step.

vasily-kirichenko: I'm afraid you are wrong. It works (I extracted the logic into a function):
```fsharp
let mergeRefs existing actual =
    // remove refs which don't actual anymore (they have been removed from the project)
    let existing =
        existing |> List.filter (fun existingRef -> List.exists ((=) existingRef) actual)
    // get refs which don't exist in the existing file
    let newExtraRefs =
        actual |> List.filter (fun actualRef -> not <| List.exists ((=) actualRef) existing)
    // concatenate old survived refs and the extra ones
    existing @ newExtraRefs

mergeRefs [3;2;1] [1;3;4]
val it : int list = [3; 1; 4]
```

dungpa: I thought results are `[3;2;1;4]`. But if preserving new entries in `existing` isn't a valid concern, then it's fine too.

PR #922, opened 2015-01-14
FsiReferenceCommand does not save script content if its not changed

vasily-kirichenko:
and: FsiReferenceCommand does not force project to save if the project has not been changed


dungpa:
Cool.

PR #926, opened 2015-01-31
Old colorization info used while a new one is calculating which results ...

vasily-kirichenko:
...with no periods of time when the code is not colorized at all


dungpa:
:+1:

Issue #931, opened 2015-02-12
[Find all refs / Rename] let bindings declared in a private module are not considered as private


vasily-kirichenko:
```fsharp
module private M =
    let getInfo () = ()
```

Find all refs on `getInfo` checks 15 projects in my solution. Maybe I don't understand private modules semantic?

vasily-kirichenko:
Fixed in https://github.com/fsprojects/VisualFSharpPowerTools/pull/932

It's funny that this bug has been unnoticed for so long time :)

PR #932, opened 2015-02-12
Find all refs and Rename searches for private / internal symbol usages i...

vasily-kirichenko:
...n projects where the symbol is declared (only)

Fix for https://github.com/fsprojects/VisualFSharpPowerTools/issues/931


OkayX6:
There are already regression tests for this, right?

vasily-kirichenko:
No.

dungpa:
The changes are relatively easy to understand. And it's not easy to add tests for this (results aren't changed; only the amount of work is reduced).

PR #933, opened 2015-02-14
Fix inconsistent coloring during editing

vasily-kirichenko:
The idea that we used to get current fresh snapshot inside `triggerClassificationChanged`, then tried to translate calculated spans to it. Now the very same snapshot on which the spans are calculated is explicitly passed to `triggerClassificationChanged`, so almost no translation is needed (for previously calculated Unused spans only) and it seems to fix the annoying inconsistent colorization during editing. 

Please, use it in your daily work and keep an eye on any colorization glitches.

### DIFF
--- a/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
+++ b/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
@@ -188,7 +188,8 @@ type SyntaxConstructClassifier
                     |> Map.ofArray
                                
                 fastState.Swap (function
-                    | FastStage.Data data -> FastStage.Data { data with Spans = spans; SingleSymbolsProjects = [] }
+                    | FastStage.Data data -> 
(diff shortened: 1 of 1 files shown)


vasily-kirichenko:
It's still produce colors for outdated snapshot. 

Guys, do you know open source classifier to look into for inspiration? I'm afraid I work with continuously changing snapshots in a wrong way, which easily results with classifier spans returned for a fresher snapshot.

vasily-kirichenko:
I found a couple of places where wrong versions of snapshots are used. I've kept an eye on the coloring today and it seems to work ok (finally).

dungpa:
Sorry, I'm on vacation at the moment. It means I can't really test it for the next 2-3 weeks.

The changes look good. I wish we create some kind of unique types for snapshots so they are impossible to be used in a wrong way. 
If you're fine with current state, I'll merge it and we find a way to improve later.

vasily-kirichenko:
OK, I'll test it a bit more for a day or two, than merge myself. 
The idea about unique types for snapshots is interesting, thanks!

PR #936, opened 2015-02-23
Operator coloring

vasily-kirichenko:
Examples:
image
image
image
image
image
image

Limitations:
* it does not colorize `=` in value definition like `let x = 1`
* it does not colorize `:>`, `:?>`, `->`, `<-` (and maybe some other build-in operators)

Default color for Operator category is the same as for plane ident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -183,3 +183,4 @@ paket-files
 *.xml
 
 release.cmd
+_NCrunch_*
--- a/src/FSharpVSPowerTools.Core/LanguageService.fs
+++ b/src/FSharpVSPowerTools.Core/LanguageService.fs
@@ -121,11 +121,13 @@ type AllowStaleResults =
     | No
 
 type WordSpan = 
-    { Line: int
+    { SymbolKind: SymbolKind
+      Line: int
       StartCol: int
       EndCol: int }
-    static member FromRange (r: Range.range) = 
-        { Line = r.StartLine
+    static member FromRange kind (r: Range.range) = 
+        { SymbolKind = kind
+          Line = r.StartLine
           StartCol = r.StartColumn 
           EndCol = r.EndColumn }
     member x.ToRange() = x.Line, x.StartCol, x.Line, x.EndCol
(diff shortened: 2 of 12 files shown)


7sharp9:
Out od interest what logic is used to determine an operator for colouring?

vasily-kirichenko:
``GetAllUsesOfAllSymbolsInFile`` -> for each `SymbolUse` get all tokens which are in its range, then apply some logic (https://github.com/vasily-kirichenko/FSharpVSPowerTools/blob/operator-coloring/src/FSharpVSPowerTools.Core/Lexer.fs#L86) to determine what kind of "symbol" it is, one of:
```fsharp
type SymbolKind =
    | Ident
    | Operator
    | GenericTypeParameter
    | StaticallyResolvedTypeParameter
    | Other
```
As a result, I highlight only those operators which `SymbolUse` is returned for.
I'm afraid it's impossible to highlight operators based on lexer only, because I see no way to exclude `<_>` in `type T<'a>() ...`, for example. 
In order to highlight `=` in `let x = 1` or `:>`, `:?>` and the like, I may get all tokens of type `TOKEN_EQUALS`, `TOKEN_GREATER`... and filter out ones which overlap with already calculated classification spans (of all kinds - types, functions, operators by current algorithm...). 

Yes, I agree that all this looks over complicated for such a simple task, so any suggestions are very welcome.

7sharp9:
Yes its fiddly, I have lots of the same sorts of issues.  

Incidentally `:?>` is a `TOKEN_COLON_QMARK_GREATER` token, and `:>` is a `TOKEN_COLON_GREATER`

vasily-kirichenko:
OK, then I'll add support for fixed number of built-in operators. May be.

vasily-kirichenko:
A small win :)

image

vasily-kirichenko:
It's ready for merging

image

dungpa:
Would you mind if I review it when I fly home on Friday?

vasily-kirichenko:
Absolutely :)

vasily-kirichenko:
* [ ] support `:?` operator

dungpa:
Out of curiosity, what is this "F# Operators" different from the default "Operator" category that comes with Visual Studio? The default 'Operator' does color `:>, :?>, ->, <-` AFAIK.

The code looks good (Sorry I was looking at this a bit late. My flight was delayed for 24h :-))

vasily-kirichenko:
Hm. That default Operator category colors what you mentions and `=`, but not custom operators. This PR colors everything.

dungpa:
Great jobs.

PR #945, opened 2015-03-05
Implement Navigate to Source

### LINKED ISSUE #201: F12 Navigate to Source
Along the same lines as [Ref12](https://github.com/SLaks/Ref12) and ReSharper's Navigate to Sources from Symbol Files, it would be great to pull up the actual source code in Visual Studio instead of just the metadata when you hit F12 if the pdb is present and it is [source linked](https://github.com/ctaggart/SourceLink). [FSharp.Data 2.0 is](http://blog.codebeside.org/blog/2014/03/10/fsharp-data-2-0-0-released/) and would be a good test. I got this project cloned and building last week. I think a lot of code can be borrowed from the Rename support.

I'm interested in helping out on this, especially from the pdb and SourceLink side of things.

dungpa:
Close #201.

User voice request: http://vfpt.uservoice.com/forums/247560-general/suggestions/5736626-f12-navigate-to-source

If both navigate to source and go to metadata are enabled, first we try to navigate to source and fall back to go to metadata if symbol source is not available.

For Navigate to Source, two things have been implemented:

 - If a symbol originates from .NET framework assemblies, open its location at http://referencesource.microsoft.com/. This is an exact port of Ref12 to F# https://visualstudiogallery.msdn.microsoft.com/f89b27c5-7d7b-4059-adde-7ccc709fa86e

 - If a symbol comes from an assembly with source-indexed pdb, try to download the relevant source file and display it. 

References:

 - http://www.lionhack.com/2014/01/14/advanced-dotnet-debugging-pdbs-and-symbols/
 - http://www.wintellect.com/devcenter/jrobbins/pdb-files-what-every-developer-must-know
 - http://blogs.msdn.com/b/sburke/archive/2008/01/16/configuring-visual-studio-to-debug-net-framework-source-code.aspx


dungpa:
Question 1: 
For the source-indexed pdb part, should we change to display links on web browsers? 

It seems that downloading source files doesn't give an advantage. The files are ad hoc and there is no way we can navigate further from them (since they have no context and no associated project). 

Question 2:
How do we improve discoverability of pdb files?

At the moment, there are very few assemblies that come with source-indexed pdb files. Most of the time, there is no pdb file at all. How should we tell users to download pdb files for improving navigation? Otherwise, there is very little value in the pdb part of the feature.

Any ideas? 
@vasily-kirichenko @OkayX6 @ctaggart @forki

ctaggart:
1. If there is no way to navigate after downloading the source file, then opening up the web browser to the exact line of the commit would be very cool. In this way, it would work just like Ref12. This would only work for well known providers like GitHub and would probably be best if this was a configurable option. 

2. It is true that few assemblies currently come with source-indexed pdb files in the NuGet Gallery. Source indexing has traditionally been difficult to do and was often done only on TFS On-Premise builds. It is used widely there. `SourceLink` aims to make this practical for everyone. If I formally release a non pre-release version of `SourceLink.exe` in addition to the `SourceLink.Fake`, I think this will increase adoption. This feature essentially makes Ref12 available for anyone who wants to source index their pdb files and will greatly increase adoption! I'll definitely be showing it off at a user group.

dungpa:
Here is a demo:

navigate_to_source

vasily-kirichenko:
F12 on `String`:

image

vasily-kirichenko:
It was loading for about 20 seconds, the proper page appeared for a second, than the homepage appeared.

vasily-kirichenko:
image

Maybe it's worth to mark such temp files somehow?

dungpa:
@vasily-kirichenko I plan to change it so that F12 goes to source code providers (GitHub, etc.) directly. We can't create fake projects like we did with Go to Metadata since there is no context and source files can be of any .NET languages (F#/C#/VB, etc).

dungpa:
@vasily-kirichenko I can't repro the behaviour with `String` symbol. Do you have an ad blocker that may intervene the behaviour of the reference source web page?

dungpa:
Ok. Now we open links from http://referencesource.microsoft.com/ and GitHub. I think the PR is ready (docs have been added).

I really appreciate if someone uses v1.7.4 from CI and gives feedback https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/1.8.0-a1503070152

vasily-kirichenko:
I've tested it on FCS (don't know any other package which supports source links).
Works ok.
About the issue with redirecting to http://referencesource.microsoft.com, I turned off both of my ad blockers and it did't help.

src/FSharpVSPowerTools.Logic/ReferenceSourceProvider.fs
```diff
@@ -0,0 +1,91 @@
...
+                Some mem.XmlDocSig
+            | TypedAstPatterns.Entity(entity, _, _) ->
+                Some entity.XmlDocSig
+            | _ -> None
+        xmlDocSig
+        |> Option.iter (fun xmlDocSig ->
+            symbol.Assembly.FileName
+            |> Option.map Path.GetFileNameWithoutExtension
+            |> Option.iter (fun assemblyName ->
+                let url = baseUrl + "/" + assemblyName + "/a.html#" + getMD5Hash xmlDocSig 16
```
vasily-kirichenko: MD5 is always 16 bytes length. So, you can remove `digits` from `getMD5Hash` and `typeArrayToHexString`.

dungpa: Thanks, done.

dungpa:
You can test FAKE (build.fsx), SourceLink and probably FSharp.Data as well.

Could you try Ref12 on C# code? I think I use the exact protocol they have there.

vasily-kirichenko:
Checkboxes layout is a bit broken 

image

vasily-kirichenko:
Tried to navigate to source on a FAKE type from our `build.fsx` https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/build.fsx#L80:

image

vasily-kirichenko:
Latest build shows the following on F12 on `Path.GetFileWithoutExtension`

image

Ref12 behaves exactly like VFPT (redirect to the home page), so it seems to be a my local issue.

vasily-kirichenko:
F12 on `IDisposable` from `fs`:
http://referencesource.microsoft.com/mscorlib/a.html#1f55292c3174123d

from `cs` (Ref12);
http://referencesource.microsoft.com/mscorlib/a.html#4e23fed29bd598fa3f28f6f3f75757e3

It seems you broken that MD5 stuff. It's 16 *bytes*, 32 chars in hex.

vasily-kirichenko:
`digits` should always be 16: `let digits = bytes.Length`.

dungpa:
I think the latest FAKE version isn't source indexed correctly /cc @forki. See a similar bug at https://github.com/fsharp/FSharp.Compiler.Service/issues/312. It works correctly for an old FAKE version (e.g. from our release branch).

@ctaggart Could you advise a better way to run source indexing at https://github.com/fsprojects/ProjectScaffold/blob/aed8942bc33c94984813390516f31dec726339e8/build.template#L153 that is not vulnerable to that FSharp.Core bug?

Issue #947, opened 2015-03-09
navigate to source for Octokit.net not working


ctaggart:
I'm testing out the latest build which has the `Navigate to Source` feature.
https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/1.8.0-a1503071249/artifacts

I created a new project with this NuGet `packages.config`:
```
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Octokit" version="0.7.2" targetFramework="net45" />
  <package id="SourceLink.Core" version="0.5.0-ci1411242108" targetFramework="net45" />
</packages>
```

The `Library1.fs`:
```
module Test

let md5 = SourceLink.Crypto.hashMD5 "a.txt"

// https://github.com/octokit/octokit.net/
let dt = Octokit.Helpers.UnixTimestampExtensions.FromUnixTime 1425868070L
let github = Octokit.GitHubClient(Octokit.ProductHeaderValue "MyAmazingApp")
let a = github.Activity.Events.GetAll()
()
```

It correctly navigates to `hashMD5` implementation which is awesome to see!

It does not work when I try to navigate to `FromUnixTime` or `GitHubClient` constructor or `GetAll`. It doesn't seem to be working for the Octokit.net library. It is source indexed.

dungpa:
Is there any error message being shown on the status bar? Or does VS just silently do nothing?

ctaggart:
If the metadata option is enabled, it goes to the metadata. If the metadata option is disabled, the status bar shows: 
![image](https://cloud.githubusercontent.com/assets/80104/6560758/bf80e868-c649-11e4-8a54-61d2be0f5612.png)
I am able to start debugging and step into `FromUnixTime` so I know it is source indexed correctly.

dungpa:
This issue is not worth fixing unless SourceLink has better support for symbols at design time.

PR #948, opened 2015-03-09
Use reflection to get source location when FCS doesn't give enough information

dungpa:
Partial fix for #947 (NOT YET READY).

For Octokit NuGet package, FCS returns faulty ranges (no range or a dummy range with wrong information). We try to use Reflection to recover source locations. I'm not sure it's a good approach; any other idea is appreciated.

@ctaggart It seems `pdbReader.GetMethod` doesn't work correctly for any overriden/abstract member e.g. any symbol from this line:

    let a = github.Activity.Events.GetAll()

Is this a SourceLink's bug? Any idea?

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -12,6 +12,7 @@ nuget FsCheck
 nuget NUnit
 nuget Foq
 nuget Nuget.CommandLine
+nuget Octokit
 nuget SourceLink.SymbolStore
 
 github fsharp/FAKE modules/Octokit/Octokit.fsx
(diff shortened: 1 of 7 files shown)


dungpa:
@ctaggart I need your help with this one to move forward. Should I wait for changes in SourceLink or discard this PR as we don't support source-indexed C# assemblies at this point?

ctaggart:
As I just mentioned in ctaggart/SourceLink#63, not all files or methods gets indexed by the compilers. F12 at design time will only work with concrete methods that get source indexed, not any of the interfaces.

dungpa:
That sounds really limiting. On a related note, can signature files (*.fsi) be source indexed?

dungpa:
Closing this PR until there is more versatile support for pdb files from SourceLink.

ctaggart:
I believe Portable PDBs https://github.com/ctaggart/SourceLink/issues/93 will solve several of the issues here. I'll create a new issue here.

PR #952, opened 2015-03-19
Remove order attributes in order that coloring is rendered in VS2012

### LINKED ISSUE #939: No syntax coloring when using dark theme
I'm using VS2012, and I don't get any syntax coloring when I use the dark theme. Well, apart from the type/let/with/etc. being blue and strings being orange. If I switch to the light or blue theme, there are no problems.

Before I installed VS2012, I installed VS2013 Community Edition, but had to uninstall it again, since it kept crashing. I don't know if this could have anything to do with it?

dungpa:
Fix #939.

Effects:

 - VS 2012: Coloring is rendered on both Dark and Light theme.
 - VS 2013: No change in behaviour. Printf coloring and escaped character coloring are shown. Debug markers appear on top of color tags.


dungpa:
@vasily-kirichenko Could you check how this change affects coloring in VS 2015?

dungpa:
We have very limited choices in controlling ordering of color tags. Any better idea to solve the problem for VS 2012 is welcomed.

vasily-kirichenko:
It works in VS 2015 as before: all categories are OK, except Printf and Escaped.

PR #955, opened 2015-03-31
Improve Active patterns' signature generation

### LINKED ISSUE #908: [Go to metadata] Missing open statements when generating metadata for active patterns
For `PropertyGet` active pattern it generates the following (this is the whole output):

```fsharp
/// An active pattern to recognize expressions that represent the read of a static or instance property, or a non-function value declared in a module
/// input: The input expression to match against.
val (|PropertyGet|_|) : Expr -> (Expr option * PropertyInfo * Expr list) option
```

where `Expr` type is unknown (I mean `open Microsoft.FSharp.Quotations` is absent).

dungpa:
Fix #908.
It requires some changes available in FCS 0.0.87.


dungpa:
Ok, done.

PR #965, opened 2015-05-10
Add SymbolInfoMargin

vasily-kirichenko:

### DIFF
--- a/src/FSharpVSPowerTools.Logic/Constants.fs
+++ b/src/FSharpVSPowerTools.Logic/Constants.fs
@@ -55,4 +55,4 @@ let [<Literal>] cmdidPreviousHighlightedReference = 2401u
 let [<Literal>] cmdidGenerateReferencesForFsi = 0x100u
 let [<Literal>] guidAddReferenceInFSICmdSetString = "8c9a49dd-2d34-4d18-905b-c557692980be"
 let guidGenerateReferencesForFsiCmdSet = Guid(guidAddReferenceInFSICmdSetString)
-    
+let [<Literal>] symbolInfoMargin = "vfpt.symbolinfo.margin";
--- a/src/FSharpVSPowerTools.Logic/FSharpVSPowerTools.Logic.fsproj
+++ b/src/FSharpVSPowerTools.Logic/FSharpVSPowerTools.Logic.fsproj
@@ -66,7 +66,7 @@
     </Resource>
     <Resource Include="MoveToFolderDialog.xaml">
     </Resource>
-    <Resource Include="SymbolInfoMargin.xaml" />
+    <Resource Include="QuickInfoMargin.xaml" />
     <Compile Include="Resource.fs">
       <Link>Common/Resource.fs</Link>
     </Compile>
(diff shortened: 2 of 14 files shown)


dungpa:
Oh, cool. You can declare yourself an Xaml expert by now :-D.

vasily-kirichenko:
;))) I'll add a setting for this in a minute.

vasily-kirichenko:
So, it shows the first line from the standard tooltip. It works ok for simple cases, but does not for records and classes (it shows `type Class =`), and for overloaded methods (it shows the first overload). 

Also it seems that XML Docs should be shown as well, but it requires a bit margin, which I don't like.

vasily-kirichenko:
done.

vasily-kirichenko:
This feature turns out not to be as useful :) The info is too far from the corresponding symbol and eyes don't like to move up and down all the time.

However, if we add more information into it, it may be more useful. I mean, XML docs (the first line of them, for example), warning and errors. Adding the latter two things would be especially cool because currently there's no way to see them without mouse.

@dungpa @dsyme Is it possible to get warnings / errors under the caret from either FCS or Visual F# (like we call the lexer state via reflection)?

dungpa:
I agree that aggregation of information would be useful.

You can add a line for XmlDoc and truncate the doc until certain length using `...`.

Regarding getting warnings/errors, it should be retrievable as `ITagAggregator<_>`.
So instead of `TextMarkerTag`, you should use `ErrorTag` or `IErrorTag` as per https://github.com/fsprojects/VisualFSharpPowerTools/blob/bc76dc272cb941fdbc04e2a0d203a685c5d3bc84/src/FSharpVSPowerTools/HighlightUsageFilterProvider.cs#L41.

You can then filter those error tags by positions (see an example at https://github.com/fsprojects/VisualFSharpPowerTools/blob/bc76dc272cb941fdbc04e2a0d203a685c5d3bc84/src/FSharpVSPowerTools.Logic/HighlightUsageFilter.fs#L49-L52).
Here is a reference to `ErrorTag` class: https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.text.tagging.errortag.aspx.

vasily-kirichenko:
Yes, duplicating tooltip in this margin is useless. But showing warnings and errors does not. 

I tried to use `ITagAggregator`. It works but error info is always `null`, i.e. error tags are indeed there, but they does not contain error details. It seems VFT tooltips are getting the details somewhere else, not from the tags themselves.

enricosada:
awesome feature @vasily-kirichenko !

dungpa:
@vasily-kirichenko Could you push the code so that I can have a look? I have hoped that error details are available in `ErrorTag.ToolTipContent`.

vasily-kirichenko:
@dungpa Take me a minute, the code is at home :) Yes, I examined 'ToolTipContent` and it's always `null`.

vasily-kirichenko:
@enricosada thanks :)

vasily-kirichenko:
@dungpa sorry, cannot reach home machine. Will push the changes tonight.

vasily-kirichenko:
@dungpa Done. I removed showing type info entirely.

dungpa:
Bummer. It doesn't work indeed. Sorry for the misleading advice.

A solution is to use `FSharpErrorInfo` from type-checked results i.e. https://github.com/fsprojects/VisualFSharpPowerTools/blob/e2b57fc184b8568304fb17ed3b1c7dbaddbb4f9f/src/FSharpVSPowerTools.Core/LanguageService.fs#L58-L61. It comes from the same source of information as VFT; they should be consistent.

Another alternative is to hijack Error Lists and capture their error messages. It is a hack though (see https://github.com/abcminiuser/marginoferror/blob/master/MarginOfError/src/Tagger/ErrorTagger.cs)

vasily-kirichenko:
Yes, hijacking Error lists is a very dirty hack (however, very fast I believe). I'm trying `FSharpErrorInfo` option.

BTW, VFPT is almost unusable on our Logic project - after 5 minutes of code changing VS starts GCing like crazy. Can you confirm?

PR #973, opened 2015-05-20
Error margin

vasily-kirichenko:
It replaces the previous PR. It does not duplicate symbol's tooltip anymore, it shows error / warning at the caret position.


vasily-kirichenko:
The interesting thing that it shows error and warnings which the tooltip does not:

dungpa:
It is a nice side effect, isn't it?

I still think it is useful to duplicate tooltip information in the margin because:

 - Tooltip information is useful for non-error cases.
 - QuickInfo tooltips aren't copyable and controllable. I presume people want this info in a persistent place and the info can be copied to clipboard.

If you worry that the margin is too big, you can display the first line of comments and one-liner signature. For example, long type signature can be shortened to `type R` with only type names.

@enricosada @OkayX6 @forki @ovatsus Do you have any ideas to chime in?

vasily-kirichenko:
Yes, the side effect is cool :)

I'm still not sure that tooltip duplication is a good idea:

* the margin is constantly updating, which is distracting 
* what should be shown in case of error? The error? The type info? Both? If both, then it's necessary to make the margin taller to fit at least two lines of text. I'm against it. 

Maybe it's better to use a tooltip-like window instead, which would displaying all the info (standard tooltip + warnings + errors)? It must be pops up by a hot key, of course (not by mouse hovering over a symbol).

dungpa:
Yeah, I meant to display both. 

So the margin consists of at most 3 lines: one for comments, one for type signatures and one for errors/warnings. Being able to copy everything to clipboard would be cool. Perhaps it's too much for a small margin. I'm still interested in what others think about this.

OkayX6:
I need to compile the PR to really test it, but I have never seen this margin used before? To me it is associated to the VsVim plugin??

What is it usually used for?

OkayX6:
But, honestly from Vasily's animation, I quite like it.
I like having errors spotted immediately.

But, I also understand that some people can be bothered or distracted. I know people who were refusing to have a dual screen at Microsoft because they would be distracted with 2 screens... A matter of choice.

dungpa:
VsVim and Viasfora both have similar controls.

Here is the context.
We would like to add a feature that provides more information on symbols under cursors. Multiple versions have been requested at:
 - http://vfpt.uservoice.com/forums/247560-general/suggestions/5737085-add-a-pane-that-constantly-shows-the-tooltip-conte
 - http://vfpt.uservoice.com/forums/247560-general/suggestions/6143128-show-type-annotation-on-expression

The PR has added error/warning information under the cursor. It's useful, no doubt.
 We are debating whether XmlDoc comments and type/function signatures should be added to the panel as well. 

Pro: more information --> more useful
Cons: the margin becomes too big --> more distracting to users

What's your thought?

vasily-kirichenko:
There are two distinct problems with large margin:

* it steals too much space from editor window
* it updates continuously

If we show errors / warnings only, it's updated very rarely, which is good.

src/FSharpVSPowerTools.Logic/Constants.fs
```diff
@@ -55,4 +55,4 @@ let [<Literal>] cmdidPreviousHighlightedReference = 2401u
 let [<Literal>] cmdidGenerateReferencesForFsi = 0x100u
 let [<Literal>] guidAddReferenceInFSICmdSetString = "8c9a49dd-2d34-4d18-905b-c557692980be"
 let guidGenerateReferencesForFsiCmdSet = Guid(guidAddReferenceInFSICmdSetString)
-    
\ No newline at end of file
+let [<Literal>] symbolInfoMargin = "vfpt.symbolinfo.margin";
```
dungpa: Change to 

```fsharp
let [<Literal>] quickInfoMargin = "vfpt.quickinfo.margin"
```

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -0,0 +1,104 @@
...
+    let docEventListener = new DocumentEventListener ([ViewChange.layoutEvent view; ViewChange.caretEvent view], 200us, updateAtCaretPosition)
+
+    interface IWpfTextViewMargin with
+        member __.VisualElement = upcast visual
+        member __.MarginSize = visual.ActualHeight + 2.
+        member __.Enabled = true
+        
+        member x.GetTextViewMargin name =
+            match name with
+            | Constants.symbolInfoMargin -> upcast x
```
dungpa: I think pattern matching always succeeds here because the literal is in lower case (see https://fslang.uservoice.com/forums/245727-f-language/suggestions/6668206-warn-when-literal-attribute-is-used-with-lowercase). Perhaps use `if/else` here instead.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.xaml
```diff
@@ -0,0 +1,13 @@
...
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
+             xmlns:sys="clr-namespace:System;assembly=mscorlib" 
+             xmlns:vs="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.11.0" 
+             xmlns:pt="clr-namespace:FSharpVSPowerTools.QuickInfo"
+             mc:Ignorable="d" 
+             Height="24"
+             d:DesignHeight="25" d:DesignWidth="600">
+    <TextBox IsEnabled="True" IsReadOnly="True" Text="{Binding QuickInfo, Mode=OneWay}" Background="{DynamicResource {x:Static SystemColors.ControlBrushKey}}" />
```
dungpa: Is there any XAML trick to make double click on the text box a select all command? Currently I have to press Ctrl + A to select all before copy the text.

dungpa:
The PR looks good. It would be great if you can add a short doc page for the new feature.

src/FSharpVSPowerTools.Logic/Constants.fs
TWith2Sugars: I think the literal's need to start with an uppercase letter: http://stackoverflow.com/questions/3890037/literal-attribute-not-working

vasily-kirichenko:
Done all. Feel free to fix the docs English.

docs/content/quickinfopanel.fsx
```diff
@@ -0,0 +1,17 @@
...
+// This block of code is omitted in the generated HTML documentation. Use 
+// it to define helpers that you do not want to show in the documentation.
+#I "../../bin"
+
+(**
+Quick Info panel
+-----------------------
+
+This panel is a small margin located at the bottom of the code editor. It automatically shows errors and warnings under the
+cursor, eliminating the need to use mouse to see tooltip with the same info.
```
dungpa: mouse --> mouses
tooltip --> tooltips

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -357,6 +357,9 @@ module String =
                | "" -> str, None
                | index -> str.Substring (0, str.Length - index.Length), Some (int index)
 
+    let trim (x: string) = match x with null -> null | x -> x.Trim()
+    let split options (separator: string[]) (x: string) = match x with null -> null | x -> x.Split(separator, options)
```
dungpa: Could you change `x` to a better name for more readable function signature :-D.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -0,0 +1,105 @@
...
+
+    let flattenLines (x: string) =
+        match x with
+        | null -> ""
+        | x ->
+            x 
+            |> String.split StringSplitOptions.RemoveEmptyEntries [|"\r\n"; "\r"; "\n"|]
+            |> Array.mapi (fun i x -> 
+                if i > 0 && x.Length > 0 && Char.IsUpper x.[0] then 
+                    ". " + (String.trim x) else " " + (String.trim x))
```
dungpa: `else` block is misaligned.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -0,0 +1,105 @@
...
+            |> String.concat ""
+            |> fun x -> if x.[x.Length - 1] <> '.' then x + "." else x
```
dungpa: Could we guard for empty strings here? As it is, the code could throw exceptions.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.xaml
TeaDrivenDev: Would this this help? http://stackoverflow.com/questions/16041556/is-there-a-way-in-xaml-to-select-all-text-in-textbox-when-double-clicked (Haven't used it myself, though.)

vasily-kirichenko: @TeaDrivenDev I've done what they suggest there, but it now crash with this:

image

Any ideas?

dungpa: /cc @ReedCopsey: any ideas?

vasily-kirichenko:
I removed that buggy xaml stuff and implement the following double click behavior, right in code:

* select all text
* copy it to clipboard 

The PR is ready for merging.

dungpa:
Cool.

Great jobs with reorganization of general dialog. Why wouldn't we think of the grouping technique before?

PR #982, opened 2015-05-26
Revert color classification defs order attribute

vasily-kirichenko:


dungpa:
Did the `StartsWith(...Ordinal)` call actually improve performance there? Do you still see a hotspot with the profiler?

vasily-kirichenko:
It shows 4%, was - 9% :) it's still red though.

PR #984, opened 2015-05-26
fixed: type constraints are not taken into account when calculating unused open declarations

vasily-kirichenko:


src/FSharpVSPowerTools.Core/OpenDeclarationsGetter.fs
```diff
@@ -348,7 +348,7 @@ module OpenDeclarationGetter =
             let res = 
                 match symbolUse.SymbolUse.Symbol with
                 | Pattern | RecordField _
-                | Entity (Class | (AbbreviatedType _ | FSharpType | ValueType | FSharpModule | Array), _, _)
+                | Entity (Class | Intf | (AbbreviatedType _ | FSharpType | ValueType | FSharpModule | Array), _, _)
```
dungpa: Why not use `Entity(_, Interface, _)` similar to `Entity(_, _, Tuple)` below?

The `Intf` active patterns seem very superficial.

vasily-kirichenko: You are right. It turns out checking for interface is not needed at all.

dungpa:
:+1:

PR #986, opened 2015-05-29
ProjectItem.FileNames use 1 based indexing

enricosada:
0 based index works for compatibility with c#, but raise a Debug.Assert dialog on visualfsharp

ref [MSDN](https://msdn.microsoft.com/en-us/library/envdte.projectitem.filenames.aspx):
> index
Type: System.Int16
Required. The index of file names from 1 to FileCount for the project item.


enricosada:
there is a better place? an helper class for DTE things? 
i'd like to do add a function `DTE.ProjectItem -> int -> string` with documentation about 1 based indexing instead of inline comment

dungpa:
We have a few extensions methods for `ProjectItem` at https://github.com/fsprojects/VisualFSharpPowerTools/blob/38715049c6c3a5b03478dd19bac47ac2c94a6382/src/FSharpVSPowerTools.Logic/VSUtils.fs#L180-L197.

enricosada:
extracted function VSUtils.filePath

dungpa:
:+1:

Issue #958, opened 2015-04-07
Syntax Coloring Latency


xenocons:
Hello,

I am not sure how to explain this issue, but I notice *significant* overhead when using Syntax Coloring, specifically, frequently when I am typing, the VS text editor 'freezes' input for roughly 1 to 2 seconds or so, my suspicion is that the cause is the Syntax Coloring (or something related to it) as when I disable it, then close the text editor and reopen it, the problem seems to go away.

vasily-kirichenko:
What profiler says?

xenocons:
Hi @vasily-kirichenko , I produced the behaviour and copied the output window:
[VFPT][Information 09:23:01 AM] [SyntaxConstructClassifier] [Fast stage] 
Total = 00:00:00.3352192
getCategoriesAndLocations, 00:00:00.3142426
GetAllUsesOfAllSymbolsInFile, 00:00:00.0194682
instance.GetAllUsesOfAllSymbolsInFile, 00:00:00.0194625
LS allSymbolsUses, 00:00:00.0155502
LS GetAllUsesOfAllSymbolsInFile, 00:00:00.0026046
ParseFileInProject, 00:00:00.0014041
LS ParseAndCheckFileInProject, 00:00:00.0012935

[VFPT][Information 09:26:29 AM] [SyntaxConstructClassifier] [Fast stage] 
Total = 00:00:00.1517972
ParseFileInProject, 00:00:00.1004051
GetAllUsesOfAllSymbolsInFile, 00:00:00.0256742
instance.GetAllUsesOfAllSymbolsInFile, 00:00:00.0256612
getCategoriesAndLocations, 00:00:00.0256038
LS GetAllUsesOfAllSymbolsInFile, 00:00:00.0212259
LS allSymbolsUses, 00:00:00.0022332
LS ParseAndCheckFileInProject, 00:00:00.0021929

I don't think these numbers represent the delay I am seeing?

vasily-kirichenko:
You are right. 
@dungpa It seems we call FCS by `Async.StartImmediately`. I tried to switch to `Async.Start` everywhere and it was crashing constantly. We should carefully examine why `StartImmediately` was used initially and eliminate all the usages where possible.

dungpa:
I think certain operations require to run on UI thread, but some uses of `StartImmediateSafe` are superficial. You can try to eliminate those constantly running on cursor changes first.

dungpa:
[...]

PR #991, opened 2015-06-04
Guard against exceptions in Syntax coloring

dungpa:
Potentially fix the exception mentioned in https://github.com/fsprojects/VisualFSharpPowerTools/issues/958#issuecomment-107026986.

### DIFF
--- a/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
+++ b/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
@@ -88,11 +88,11 @@ module private StringCategorizers =
             [r.StartLine..r.EndLine]
             |> List.map (fun line ->
                 let lineStr = getTextLine (line - 1)
-                if lineStr.Length > r.StartColumn + 1 && line = r.StartLine && line = r.EndLine && r.StartColumn + 1 <= r.EndColumn - 1 then
+                if line = r.StartLine && line = r.EndLine && r.StartColumn + 1 <= r.EndColumn - 1 && lineStr.Length > r.EndColumn - 1 then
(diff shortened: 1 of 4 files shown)


src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
```diff
@@ -88,11 +88,11 @@ module private StringCategorizers =
...
-                if lineStr.Length > r.StartColumn + 1 && line = r.StartLine && line = r.EndLine && r.StartColumn + 1 <= r.EndColumn - 1 then
+                if line = r.StartLine && line = r.EndLine && r.StartColumn + 1 <= r.EndColumn - 1 && lineStr.Length > r.EndColumn - 1 then
                     lineStr.[r.StartColumn + 1 .. r.EndColumn - 1], line, r.StartColumn + 1
-                elif lineStr.Length > r.StartColumn + 1 && line = r.StartLine then 
+                elif line = r.StartLine && lineStr.Length > r.StartColumn + 1 then 
                     lineStr.[r.StartColumn + 1 ..], line, r.StartColumn + 1
-                elif lineStr.Length > r.EndColumn - 1 && line = r.EndLine then
+                elif line = r.EndLine && lineStr.Length > r.EndColumn - 1 then
                     lineStr.[..r.EndColumn - 1], line, 0
                 else lineStr, line, 0)
```
vasily-kirichenko: This `if..elif..elif` chain is really complicated. I've no idea how to simplify it. Active patterns may be?

dungpa: I don't think it helps much. I particularly hate the partial nature of those indexers. If only they return option or choice values...

PR #995, opened 2015-06-07
QuickInfoMargin shows tooltip copy if no errors or warnings and "F# QuickInfo panel" text if neither of above.

vasily-kirichenko:


dungpa:
Cool.

Now we have tooltips in the margin very often. Do you think we need the "F# QuickInfo panel" label anymore?

I think single-line tooltip info works quite well for values, members, etc. However, tooltips displayed in QuickInfo margin are often multi-lines. When I select the text, it's weird since the last line of the tooltips appears.
If you want to make it truly single-line, we can post process e.g.

---

module A
from B

--> module A

---

type A =
  | B
  | C

--> type A = ...

---

let x: int -> int

Fullname:  A.x

--> let x: int -> int

---

What do you think?

vasily-kirichenko:
I agree "F# QuickInfo panel" should be removed.

What about tooltip post processing, I'm not sure what you mean. I've used it for a couple of days and have never seen any strange output. Could you provide more examples like "current vs desired"?

dungpa:
Before selecting text, first line of a tooltip is shown:
image

After selecting text in the margin, last line of a tooltip is shown:
image

In short, muti-line tooltips are displayed in a single-line textbox. I thought it's better to trim the tooltips to be single-line.

vasily-kirichenko:
Done.

dungpa:
Thanks.

PR #1001, opened 2015-06-10
Optimization

vasily-kirichenko:
WIP

### DIFF
--- a/src/FSharpVSPowerTools.Core/Utils.fs
+++ b/src/FSharpVSPowerTools.Core/Utils.fs
@@ -29,9 +29,15 @@ module Array =
 
     /// Returns true if one array has trailing elements equal to another's.
     let endsWith (suffix: _ []) (whole: _ []) =
-        whole 
-        |> Array.rev
-        |> startsWith (Array.rev suffix)
+        let rec loop step =
+            if step > suffix.Length then true
+            elif step > whole.Length then false
+            elif suffix.[suffix.Length - step] = whole.[whole.Length - step] then loop (step + 1)
+            else false
+
+        if suffix.Length = 0 then true
+        elif whole.Length < suffix.Length then false
+        else loop 1
 
     /// Returns a new array with an element replaced with a given value.
     let replace index value (arr: _ []) =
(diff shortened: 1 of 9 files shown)


src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -29,9 +29,15 @@ module Array =
 
     /// Returns true if one array has trailing elements equal to another's.
     let endsWith (suffix: _ []) (whole: _ []) =
-        whole 
-        |> Array.rev
-        |> startsWith (Array.rev suffix)
+        let rec loop step =
+            if step > suffix.Length then true
+            elif step > whole.Length then false
+            elif suffix.[suffix.Length - step] = whole.[whole.Length - step] then loop (step + 1)
```
dungpa: It's small thing but you can hoist `suffix.Length` and `whole.Length` to values outside the `loop` function.

src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
```diff
@@ -132,7 +132,7 @@ type SyntaxConstructClassifier
                 let tooltip =
                     vsLanguageService.GetOpenDeclarationTooltip(
                         line, endCol, lineStr, Array.toList idents, project, textDocument.FilePath, source)
-                    |> Async.RunSynchronously
+                    |> Async.RunSynchronously // todo remove RunSynchronous !!!!!!!!!!!!!!!!!!!!!
```
dungpa: I would love to see this 'todo' happens!!!

vasily-kirichenko:
When I profiled, nothing VFPT / FCS related was executed in UI thread :( It must be blocked waiting for a background thread to complete ==> UI freeze.

vasily-kirichenko:
Done :)

vasily-kirichenko:
Although it does not freeze in my scenario anymore, code editing is not absolutely smooth (as in Xamarin, for example).

dungpa:
I doubt we will ever make unused (open) declarations fast enough. Moreover, FSharpBinding in Xamarin Studio didn't have *incremental* multi-project checking last time I checked.

vasily-kirichenko:
Scenario:

* Turn on "Resolve unopened namespaces" and both "Gray out..." (I have not tried with these options off though) 
* Add `Seq.filter (fun x -> Thread )` into this pipe https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs#L119
* Wait until "add open declaration" smart tag appears 
* Play with `Thread` word a bit (delete it, type it again and so on)
* After 30 seconds of this game UI frizzed for very long time, now the lags is about 1 second or less


In short, just try to edit any error-squizzled code (preferably a large area of such code) in a not-so-small project.

PR #1003, opened 2015-06-10
Improve union case generation feature

### LINKED ISSUE #970: 'Generate union cases' injects text to the wrong place
gencase

Not 100% repro, might be a race condition where cursor is out of sync with where VFPT thinks it is

Repro in VS 2013 and 2015

### LINKED ISSUE #971: "Generate union cases" does not handle nested patterns
```fsharp
type A = A1 | A2

type B = B1 | B2 of A

let test x =
    match x with
    | B1 -> ()
    | B2(A1 | A2) -> ()
```

Hover over `A1` or `A2` in bottom case, it will give bogus suggestion/completion:

unioncase1

dungpa:
Fix #971 and add a test case.

Fix #970 and test manually that it no longer happens. It would be great if someone can verify that.

### DIFF
--- a/src/FSharpVSPowerTools.Core/UnionPatternMatchCaseGenerator.fs
+++ b/src/FSharpVSPowerTools.Core/UnionPatternMatchCaseGenerator.fs
@@ -34,24 +34,28 @@ type private Context = {
     Qualifier: string option
 }
 
-let private clauseIsCandidateForCodeGen (clause: SynMatchClause) =
+let private clauseIsCandidateForCodeGen (cursorPos: pos) (clause: SynMatchClause) =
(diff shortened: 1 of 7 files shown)


src/FSharpVSPowerTools.Core/UnionPatternMatchCaseGenerator.fs
```diff
@@ -555,7 +561,7 @@ let checkThatPatternMatchExprEndsWithCompleteClause (expr: PatternMatchExpr) =
 
             rhsExprExists && not (synExprContainsError expr.Expr)
 
-        | _ -> true && not (synExprContainsError expr.Expr)
```
vasily-kirichenko: Wow :)

PR #1008, opened 2015-06-17
Add more docs

dungpa:
My experience is that when we document something, the related issues tend to disappear. Let's talk about performance :-D.


vasily-kirichenko:
Great stuff ;)

PR #1010, opened 2015-06-17
Ensure to draw unused markers in the range of scroll bars

### LINKED ISSUE #985: Unused markers in the scrollbar area are drawn over the arrows and not redrawn on resize
Unused markers are erroneously drawn over the full range of the scroll bar, including the up and down arrow:
image

Additionally, their range is not recomputed when resizing the window, so they end up being drawn completely at the wrong place. For example this is what I got when shrinking the above window:
image

dungpa:
Fix #985.


dungpa:
@vasily-kirichenko Can you check this first? Thanks.

vasily-kirichenko:
I've checked it carefully but forget to merge :( Sorry.

PR #1012, opened 2015-06-17
Improve Unused Declaration Margin

dungpa:
Read unused declaration tags through a separate channel. It seems to work; I'm not sure about performance though.


vasily-kirichenko:
You are a genius :) Very elegant solution. Works OK for the first glance. Will use it tomorrow at work and report if the exception is gone. Thanks a lot!

dungpa:
I haven't added any cache for the unused declaration margin. I hope we aren't going to need it.

vasily-kirichenko:
Yes, I agree that caching is not needed here.

PR #1013, opened 2015-06-17
decrease FSharpChecker's internal cache to 50 projects

vasily-kirichenko:
With old value `200` when I found all refs of a symbol used in ~90 projects, VS occupied > 2GB of RAM  and became unusable due to continuous GCing. 

With the cache of size `50` subsequent launches of "Find all refs" are as slow as the very first one, but VS private set stays at ~1.6GB, which is acceptable.

### DIFF
--- a/src/FSharpVSPowerTools.Core/LanguageService.fs
+++ b/src/FSharpVSPowerTools.Core/LanguageService.fs
@@ -158,7 +158,7 @@ type LanguageService (?fileSystem: IFileSystem) =
   // when its view of the prior-typechecking-state of the start of a file has changed, for example
   // when the background typechecker has "caught up" after some other file has been changed, 
   // and its time to re-typecheck the current file.
-  let checkerInstance = FSharpChecker.Create (projectCacheSize = 200, keepAllBackgroundResolutions = false)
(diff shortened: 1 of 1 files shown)


dsyme:
Is there any way we can make the cache more adaptive to memory usage rather than using a fixed number?

dungpa:
@dsyme I'm not sure how we can do it. We have many cache layers in VFPT, memory usage is a bit unpredictable. Moreover, only a single `FSharpChecker` instance is used once it is created.

vasily-kirichenko:
@dsyme About adaptive cache. It would be great if `FSharpChecker` extends the ctor like this:

```fsharp
let checker = 
    FSharpChecker.Create (
        projectCacheSize = 50, 
        keepAllBackgroundResolutions = false, 
        isMemoryPressure = fun _ -> get_current_process_private_set() > 2GB)
```
It could call the callback before adding a new entry to the cache and, if it return `true`, remove the oldest / less recent used entry. However, if new entries hold more memory than old ones, then this algorithm won't work well. FCS could repeat calling the callback + `GC.Collect()` until the callback returns `false`, but I suspect that `GC.Collect()` may take several seconds or more.

Issue #1014, opened 2015-06-18
[Coloring] Highlighting unused declarations / open statements stops working


vasily-kirichenko:
At some point Unused stops working - the margin is not updated (it has outdated markers), new unused symbols are not highlighted as Unused.

Debugger showed that background compilation is triggered OK https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L329 but the callback https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L339 is not, hence the issue. 

I'm struggling to understand why FSC stopped calling the callback? Maybe the callback throws an exception and FSC classified it as "malicious" and removed from its list?

This never happened before your yesterday optimization.

dungpa:
The callback at https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L339 is only called when the project is compiled.

Is the `updateUnusedDeclarations` function called? We hook it at `ProjectChecked` event at https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L357.

I'm not sure why it stops working at all.

vasily-kirichenko:
Yes, I was wrong, I mean this callback https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L357

It's not called. I try to add `protect` for the whole body of it and will wait for exceptions.

vasily-kirichenko:
It stopped working again. The only errors in our log are (4 duplicates):
```
[VFPT][Error 01:18:15 ] Context: Exception occurs while querying lexing states.
Exception Message: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
Stack Trace:    at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at Microsoft.VisualStudio.Editor.Implementation.LanguageServiceClassificationTagger.TryGetStartingStateForLine(Int32 lineNo)
[...]

PR #1015, opened 2015-06-18
Possible fix for #1014

dungpa:
I can't think of any reason rather than a threading issue. Could you try this incremented build to see whether the exceptions still occur?


vasily-kirichenko:
Thanks, will try.

vasily-kirichenko:
No error so far. Great job.

Issue #492, opened 2014-06-12
[Highlight references] Does not appear after renaming


vasily-kirichenko:
![6677](https://cloud.githubusercontent.com/assets/873919/3255077/d4745e18-f204-11e3-8896-5b9905d02a92.gif)

Issue #1016, opened 2015-06-18
FSharpChecker.ProjectChecked event is raised for no reason


vasily-kirichenko:
This lambda https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L358-L375 is called 1-5 times for the same project and document even though the document has not changed - just move the caret with keyboard or click anywhere with mouse. It's extremely strange and I've no idea what's happening. 

(`StartBackgroundCompile ` here https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L329 is not called too often, it's called only if a file has changed).

dungpa:
Is the anonymous function called for different `projectFileName`? We have another `StartBackgroundCompile` inside its body https://github.com/fsprojects/VisualFSharpPowerTools/blob/f6fffb6e9d1cab23346171f675a882f3850e0e35/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs#L370.

vasily-kirichenko:
`projectFileName` is the same.

vasily-kirichenko:
I think it's triggered by this line https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Core/LanguageService.fs#L320

vasily-kirichenko:
Yes. That's because we always pass `AllowStaleResults.No`. Do you remember why?

vasily-kirichenko:
Fixed in https://github.com/fsprojects/VisualFSharpPowerTools/pull/1017

PR #1017, opened 2015-06-18
use AllowStaleResults.MatchingSource everywhere except RenameCommandFilter

vasily-kirichenko:
Fix for https://github.com/fsprojects/VisualFSharpPowerTools/issues/1016

### DIFF
--- a/src/FSharpVSPowerTools.Core/LanguageService.fs
+++ b/src/FSharpVSPowerTools.Core/LanguageService.fs
@@ -537,7 +537,7 @@ type LanguageService (?fileSystem: IFileSystem) =
 
     member x.GetAllEntitiesInProjectAndReferencedAssemblies (projectOptions: FSharpProjectOptions, fileName, source, ?withCache) =
         async {
-            let! checkResults = x.ParseAndCheckFileInProject (projectOptions, fileName, source, AllowStaleResults.No)
+            let! checkResults = x.ParseAndCheckFileInProject (projectOptions, fileName, source, AllowStaleResults.MatchingSource)
             return 
                 Some [ match checkResults.GetPartialAssemblySignature() with
                        | Some signature -> 
@@ -567,7 +567,7 @@ type LanguageService (?fileSystem: IFileSystem) =
 
     member x.GetIdentTooltip (line, colAtEndOfNames, lineStr, names, project: FSharpProjectOptions, file, source) =
         async {
-            let! checkResults = x.ParseAndCheckFileInProject (project, file, source, AllowStaleResults.No)
+            let! checkResults = x.ParseAndCheckFileInProject (project, file, source, AllowStaleResults.MatchingSource)
             return! checkResults.GetIdentTooltip (line, colAtEndOfNames, lineStr, names)
         }
 
--- a/src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
+++ b/src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
@@ -137,7 +137,7 @@ type GoToDefinitionFilter(textDocument: ITextDocument,
             let! symbolUses = 
                 vsLanguageService.GetAllUsesOfAllSymbolsInSourceString(
                     signature, filePath, signatureProject, 
-                    AllowStaleResults.No, checkForUnusedOpens=false, profiler=Profiler())
+                    AllowStaleResults.MatchingSource, checkForUnusedOpens=false, profiler=Profiler())
 
             /// Try to reconstruct fully qualified name for the purpose of matching symbols
             let rec tryGetFullyQualifiedName (symbol: FSharpSymbol) = 
--- a/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs
+++ b/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs
@@ -57,7 +57,7 @@ type HighlightUsageTagger(textDocument: ITextDocument,
         async {
             if currentRequest = requestedPoint then
                 try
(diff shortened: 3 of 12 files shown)


vasily-kirichenko:
I have no idea why some of the integration tests failed.

dungpa:
What's the goal of this PR? From which feature is the background compile triggered https://github.com/fsprojects/VisualFSharpPowerTools/blob/1b8ccb7de173bd3fe73292bc17b89edd38f0d067/src/FSharpVSPowerTools.Core/LanguageService.fs#L320?

Do you observe any performance improvement with this PR?

A bit of history:

 - We started using `AllowStaleResults.No` for `HighlightUsageTagger` in https://github.com/fsprojects/VisualFSharpPowerTools/pull/502 in order to fix https://github.com/fsprojects/VisualFSharpPowerTools/issues/492
 - We started using `AllowStaleResults.No` for `SyntaxConstructClassifier` at https://github.com/fsprojects/VisualFSharpPowerTools/pull/496

dungpa:
Probably these tests failed because we share the same file names for multiple tests. When `AllowStaleResults.MatchingSource` is used, staled results are reused and consequently wrong results are asserted.

Perhaps creating file names locally to the tests could fix it https://github.com/fsprojects/VisualFSharpPowerTools/blob/1b8ccb7de173bd3fe73292bc17b89edd38f0d067/tests/FSharpVSPowerTools.Tests/HighlightUsageTaggerTests.fs#L47.

vasily-kirichenko:
Yes, selecting text and moving cursor with keyboard seems to work faster with this PR.
Background compile should not be triggered if the file has not been changed, shouldn't it? So, only colorization and renaming should use `AllowStaleResults.No`. 

https://github.com/fsprojects/VisualFSharpPowerTools/issues/492 is not reproduced with this PR.
`SyntaxConstructClassifier` does use `AllowStaleResults.No` in this PR.

> When AllowStaleResults.MatchingSource is used, staled results are reused and consequently wrong results are asserted.

I believe `AllowStaleResults.MatchingSource` allows reuse of results if _file content_ has not been changed, not file name. Am I wrong?

dungpa:
> Background compile should not be triggered if the file has not been changed, shouldn't it? So, only colorization and renaming should use AllowStaleResults.No.

Agreed.

> I believe AllowStaleResults.MatchingSource allows reuse of results if file content has not been changed, not file name. Am I wrong?

You're right. But the integration tests use buffer contents; file names are just placeholders (they don't even exist). I think using unique file names is an OK trade-off.

vasily-kirichenko:
I cannot run tests at home

image

will fix them tomorrow.

vasily-kirichenko:
You are right, generating unique file name for each test solved the problem, except one single test: "should generate highlight usage tags for multi-project symbols". Any ideas?

vasily-kirichenko:
```
FSharpVSPowerTools.Tests.HighlightUsageTaggerTaggerTests.should generate highlight usage tags for multi-project symbols:
  Expected: [(5, 22, 5, 28); (6, 22, 6, 28)]
Actual: []
  Expected and actual are both <Microsoft.FSharp.Collections.FSharpList`1[System.Tuple`4[System.Int32,System.Int32,System.Int32,System.Int32]]>
  Values differ at index [0]
```

dungpa:
That test is a bit subtle. You can write content from https://github.com/fsprojects/VisualFSharpPowerTools/blob/7447e6792ae0ced520842caccbd592f48fce3dd9/tests/FSharpVSPowerTools.Tests/HighlightUsageTaggerTests.fs#L196-L202 to the file before executing. The goal is that its content on disk should be consistent with buffer content https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/tests/data/MultiProjects/Project2/Project21.fs.

Sorry for the inconsistence. It was passing so I never saw it as a bad practice.

vasily-kirichenko:
I added writing the content to the file, but it has not helped. It fails even being run in isolation.

dungpa:
I don't really understand why those multi-project tests are so flaky.

You can ignore it for now. It's too fragile so I always disable it on AppVeyor.

vasily-kirichenko:
OK, so we are waiting for AppVeyor to finish the build.

vasily-kirichenko:
Oops. I've just found that Highlight references does not work nicely in this PR. After renaming it does not re-highlight the symbol. After manual renaming it does not highlight the symbol at all, even after the caret has moved.

dungpa:
Is this the only issue? I think it can be fixed by checking whether the buffer has changed and using `AllowStaleResults.No` instead of `MatchingSource` for such scenarios https://github.com/fsprojects/VisualFSharpPowerTools/blob/07637552a5f2f7aafef29a3fc35149bdb153beee/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs#L95.

vasily-kirichenko:
I've tested it with the version before this PR was merged and highlighting after renaming does not work either.

vasily-kirichenko:
It's reproduced on large files in a real solution only. On a two-lines file in a toy solution everything works OK, even on this PR.

vasily-kirichenko:
1. This https://github.com/fsprojects/VisualFSharpPowerTools/blob/07637552a5f2f7aafef29a3fc35149bdb153beee/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs#L60 returns None after renaming.

2. This https://github.com/fsprojects/VisualFSharpPowerTools/blob/07637552a5f2f7aafef29a3fc35149bdb153beee/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs#L81 is call exactly twice when the caret is moved to a new symbol (including the rename scenario). wtf?

PR #1020, opened 2015-06-19
Use EditorUtils for HightlightUsageTagger

vasily-kirichenko:


vasily-kirichenko:
@dungpa please review. It does create a single Tagger for each document.

dungpa:
Why closing? I wonder whether we really need to depend on `EditorUtils`. I think use the same mechanism as https://github.com/fsprojects/VisualFSharpPowerTools/blob/ea13190c96b9c6884ed94eabb837716216578acd/src/FSharpVSPowerTools/SyntaxConstructClassifierProvider.cs#L474-L476 would work.

vasily-kirichenko:
Yeah, found it ten minutes ago :) https://github.com/fsprojects/VisualFSharpPowerTools/pull/1021

PR #1022, opened 2015-06-19
Revert "use AllowStaleResults.MatchingSource everywhere"

vasily-kirichenko:
Colorization is buggy as hell. I think all these troubles are not worth the small (if any) performance boost.


dungpa:
Yessss. :+1:

vasily-kirichenko:
:))

Issue #1023, opened 2015-06-23
Settings pages are badly layed out especially on High DPI


xperiandri:
General page of settings is cut at the left and bottom if Options window has minimum allowed size.
See screenshot. Bottom and right lines of groupbox are absent while top and left has some undesired margin
![f general page](https://cloud.githubusercontent.com/assets/2365592/8303612/c6bb4b62-19a7-11e5-8c95-7b0bbe6b8d82.png)

While Code Generation page is badly aligned at the left, top and is not resizable. Maybe it must not be resizable, but if you do not align it to the very left it looks like it should.
![f code generation](https://cloud.githubusercontent.com/assets/2365592/8303453/b7af4020-19a6-11e5-990a-86e0caa35635.png)

The line of goupbox in both mentioned pages must be aligned to a center of searchbox lens circle.
For reference look at Environment section General page.

vasily-kirichenko:
We are not UI experts and very much appreciate a help from one :) The dialog is a WinForms control, so it's easily editable.

xperiandri:
Fixed myself
Please merge

vasily-kirichenko:
Fixed in https://github.com/fsprojects/VisualFSharpPowerTools/pull/1024

PR #1024, opened 2015-06-23
Optimized layout of options pages according to Visual Studio style

xperiandri:
Fix for #1023


vasily-kirichenko:
The dialog looks great now. Thanks!

PR #1028, opened 2015-06-26
Remove '=' suffix on signatures of types	in QuickInfo margin

dungpa:
I find the `=` suffix in signatures e.g. `type A = ` very annoying, so I remove it.


vasily-kirichenko:
Great.

PR #1029, opened 2015-06-26
Add a new option for interface member identifiers

dungpa:
UserVoice request: http://vfpt.uservoice.com/forums/247560-general/suggestions/8557105-allow-renaming-of-self-reference-for-implementing

Again I flex my "great" UI skills. Any suggestion is welcome.


vasily-kirichenko:
If the identifier is empty, then the result is as following:

image

dungpa:
Ok, invalid identifiers are now disallowed.

dungpa:
Test failures are fixed separately in #1031.

vasily-kirichenko:
```fsharp
class ==> ``class`` (ok)
``class`` ==> ``class`` (ok)
``->`` ==> ``->`` (ok)
-> ==> validation error (ok)
(empty string) ==> no validation error, same behavior as before
x y ==> validation error (should be ``x y``)
```

dungpa:
Ok, done.

vasily-kirichenko:
Great. As a (arguably useful) side effect we got to know that the following compiles :)

image

PR #1034, opened 2015-07-03
Lint

vasily-kirichenko:


dungpa:
Questions:

1. Do we want to show lint warnings on VS Error List?

2. How should we manage FSharpLint option? Are they scoped for the whole Visual Studio or scoped for solutions/projects?

3. It would be nice to display colored tooltips to distinguish between VFT and VFPT tooltips. The trick is documented at http://winterdom.com/2014/01/pretty-quickinfo-tooltips-and-vs-themes

@duckmatt You advices on integrating FSharpLint are really appreciated.

vasily-kirichenko:
1. I don't think we should analyze whole projects and solutions and show hints in Error list. That's what FSharpLint does AFAIK.
2. I hope we can just add a standard property grid into settings dialog and map FSharpLint settings into it directly. 
3. I'll try to make the tooltips colored :)

duckmatt:
Thanks for looking to integrate it, it looks great.

The rules in the configuration are stored in a `Map`. Via the API you can pass in your own configuration but it's also possible to load a configuration file for a project - this traverses from the root directory of the drive down to the project directory, and for each directory if a config file is found it is applied with the rules in the child directory overwriting any from a parent directory.

The loading of a config for a project follows the same approach as StyleCop. The integration of StyleCop in visual studio provides a ui for the config but writes what you select through the ui out to a config file for the project. This approach works quite well because then the config for projects and solutions can be on the project's source control

Currently there's nothing for writing the config to a file (only reading) but this is something I could look at getting into the API as soon as possible

vasily-kirichenko:
About distinguishing from VFT, I added bold "Lint:" prefix to each hint:

image

image

About coloring, I'm not sure orange would be readable. What do you think?

JohanLarsson:
Would it make sense with a vertical squiggle for the 21 lines warning?
About colour: is there a brush in the theme that can be used?

dungpa:
@vasily-kirichenko We have the bits reading brush keys at https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/VSColors.fs. Perhaps we should reuse that instead of implementing once again.

About tag colors, why don't we use the usual blue warning colors? VS Spell Checker extension uses purple, which is quite easy to spot.

We should add a "Lint settings..." menu to project contexts. It should open a property grid (powered by FsXaml?) to populate FSharpLint options.

dungpa:
@duckmatt We really need a write API for FSharpLint configs.

 - Does FSharpLint have unique identifiers for rules? Some things like `FS0001` error idents in F# to differentiate errors/warnings. 
 - Do FSharpLint configs allow to specify ignored files in an F# project?

vasily-kirichenko:
I'd like to separate linter warnings from compiler warnings since 1. The latter are really hints, not warnings 2. There will be a lot of them. For the second season I also like the tags color be low contrast, even lighter than Orange. Orange is used for linter hints for Haskell code in Eclipse and Sublime, for example. AFAIK it also used by R# (it uses cut green tags for low priority hints though).

About setting, what is the point in per project setting? Do you have a concrete example where different settings are desirable? I'd like to have solution wide settings (stored in a file).

vasily-kirichenko:
Linter has quite useful rules:

image

image

duckmatt:
The names of the rules combined with the names of the analysers can be used to uniquely identify them, similar ids  to `FS0001` could be added if that will be needed maybe with a prefix of FL `FL0001`

There's a section in the config for ignoring files, [the default config contains assemblyinfo.*](https://github.com/fsprojects/FSharpLint/blob/master/src/FSharpLint.Framework/DefaultConfiguration.FSharpLint)

Solution wide settings are doable as long as the solution file is in a parent directory to all of the projects (would be an odd custom layout for that not to be the case). I can't think of a great example of per project settings in a solution, but perhaps there's a legacy project in a solution that you'd want laxer rules applied to than a new project being added to the same solution. I do think you'd mostly want to be setting the config for the entire solution as you say

vasily-kirichenko:
We are not going to run linter on entire projects and solutions. It works on the active open file only, so an ignoring file list has no sense.

dungpa:
Yeah, FSharpLint configs for whole solutions would be enough.

Current color looks a bit weird in dark themes. I think the old Orange color is probably better.

image

vasily-kirichenko:
It's customizable in Fonts and Colors dialog. I tried lighted Blue, see https://github.com/fsprojects/VisualFSharpPowerTools/pull/1034#issuecomment-118607162

JohanLarsson:
Is a vertical line hard to implement or bad for other reasons?

vasily-kirichenko:
@JohanLarsson show me what it looks like in other tools (R#, etc) or editors (IDEA, Eclipse, Atom, emacs, vim, sublime).

src/FSharpVSPowerTools.Logic/LintTagger.fs
```diff
@@ -0,0 +1,95 @@
...
+            let spans = 
+                match spans with 
+                | Some x -> x
+                | None -> []
+            wordSpans <- spans
+            let span = SnapshotSpan(buffer.CurrentSnapshot, 0, buffer.CurrentSnapshot.Length)
+            tagsChanged.Trigger(self, SnapshotSpanEventArgs span))
+        |> Async.StartInThreadPoolSafe
+
+    let docEventListener = new DocumentEventListener ([ViewChange.bufferEvent textDocument.TextBuffer], 200us, updateAtCaretPosition)
```
vasily-kirichenko: Buffer event is not raised. I have no idea why. The code is identical to what's used in `SyntaxConstructClassifier`. Ideas?

JohanLarsson:
I'm not I have seen it in any other tool. Thinking something like:
vertical squiggle

vasily-kirichenko:
@JohanLarsson As I remember, R# shows hints as a little line, which does not underline all the problem code. I mean it always has fixed length, very small. Could you make a screenshot?

JohanLarsson:
A bunch of screenshots:
two
random stuff

vasily-kirichenko:
Thanks. Yes, I mean then little thing which underlines the very first `string`.

src/FSharpVSPowerTools.Logic/LintQuickInfoSource.fs
```diff
@@ -0,0 +1,57 @@
...
+        textBlock.SetResourceReference(TextBlock.ForegroundProperty, VSColors.ToolTipTextBrushKey)
+        upcast textBlock
+
+    interface IQuickInfoSource with
+        member __.AugmentQuickInfoSession (session: IQuickInfoSession, quickInfoContent: IList<obj>, 
+                                           applicableToSpan: byref<ITrackingSpan>) = 
+            match session.GetTriggerPoint buffer.CurrentSnapshot |> Option.ofNullable with
+            | None -> ()
+            | Some point ->
+                use tagAggregator = tagAggregatorService.CreateTagAggregator<LintTag> session.TextView
```
dungpa: I think the reason event subscription doesn't work is that the tagger is disposed in this function when `tagAggregator` is out of scope. I think we should create a `tagAggregator` once in the provider and only dispose once it at the end. We did it that way in other features.

src/FSharpVSPowerTools.Logic/LintQuickInfoSource.fs
```diff
@@ -0,0 +1,57 @@
...
+
+                match res with
+                | [] -> ()
+                | (span, _) :: _ ->
+                    applicableToSpan <- buffer.CurrentSnapshot.CreateTrackingSpan (span.Span, SpanTrackingMode.EdgeExclusive)
+                    res |> List.map snd |> createInfoText |> quickInfoContent.Add
+        
+        member x.Dispose(): unit = 
+            if not disposed then
+                GC.SuppressFinalize x
```
dungpa: `SuppressFinalize` should not be needed. We should dispose the tag aggregator here.

src/FSharpVSPowerTools.Logic/LintTagger.fs
```diff
@@ -0,0 +1,105 @@
...
+                    if currentSnapshot = (snd wordSpans.[0]).Snapshot then
+                        wordSpans
+                    else
+                        // If the requested snapshot isn't the same as the one our words are on, translate our spans
+                        // to the expected snapshot
+                        wordSpans |> List.map (fun (warn, span) -> warn, span.TranslateTo(currentSnapshot, SpanTrackingMode.EdgeExclusive))
+                
+                for warn, span in wordSpans do
+                    if spans.OverlapsWith span then
+                        yield upcast TagSpan<LintTag>(span, LintTag(warn.Info))
```
dungpa: I think you can manipulate `spans` here if you would like to display a sub range of what FSharpLint gives.

src/FSharpVSPowerTools/LintQuickInfoProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+    [Order(Before = "Default Quick Info Presenter")]
+    [ContentType("F#")]
+    internal class LintQuickInfoProvider : IQuickInfoSourceProvider
+    {
+        [Import]
+        internal IViewTagAggregatorFactoryService viewTagAggregatorFactoryService = null;
+
+        public IQuickInfoSource TryCreateQuickInfoSource(ITextBuffer textBuffer)
+        {
+            return new LintQuickInfoSource(textBuffer, viewTagAggregatorFactoryService);
```
dungpa: We should pass tag aggregator instead of the service.

src/FSharpVSPowerTools/LintQuickInfoProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+    [Name("F# Lint Quick Info Provider")]
+    [Order(Before = "Default Quick Info Presenter")]
+    [ContentType("F#")]
+    internal class LintQuickInfoProvider : IQuickInfoSourceProvider
+    {
+        [Import]
+        internal IViewTagAggregatorFactoryService viewTagAggregatorFactoryService = null;
+
+        public IQuickInfoSource TryCreateQuickInfoSource(ITextBuffer textBuffer)
+        {
```
dungpa: I think an `IQuickInfoSource` should only be created if `LinterEnabled` is true. We should check for that option here, similar to what have been done in `LintTaggerProvider`.

src/FSharpVSPowerTools/LintQuickInfoProvider.cs
vasily-kirichenko: It requires `ITextView` which we don't have at this point. So, I have to pass the service and retrieve/dispose an `Aggregator` each time tooltip is queried.

src/FSharpVSPowerTools/LintQuickInfoProvider.cs
vasily-kirichenko: Will fix.

src/FSharpVSPowerTools.Logic/LintQuickInfoSource.fs
vasily-kirichenko: Yes, if we had a single aggregator. We dont.

src/FSharpVSPowerTools/LintQuickInfoProvider.cs
dungpa: There is `IBufferTagAggregatorFactoryService` to be use with `ITextBuffer`. Oh, dark magic :(.

dungpa:
I think the key point is to use `IBufferTagAggregatorFactoryService` https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.text.tagging.ibuffertagaggregatorfactoryservice.aspx. Everything should be easy to handle if we have a single aggregator.

vasily-kirichenko:
Yeeeah. `IBufferTagAggregatorFactoryService`. Now everything looks perfect and even the buffer changed event is raised, but `tagAggregator.GetTags span` returns empty seq :(

vasily-kirichenko:
I limit the tags length to max 2 characters, the result is like R# does, much less disturbing:

image

duckmatt:
@vasily-kirichenko that looks good, looks like it also resolves https://github.com/fsprojects/FSharpLint/issues/102

vasily-kirichenko:
Found an exactly same code https://github.com/tunnelvisionlabs/LangSvcV2/blob/7b246e1499664e0e795f472a503c15cbb0919fff/Tvl.VisualStudio.Language.Antlr3/AntlrQuickInfoSourceProvider.cs 

No idea why my don't work.

vasily-kirichenko:
@duckmatt I just added `Seq.distinctBy (tooltip, range)`

vasily-kirichenko:
It almost works. it's dirty. I've no idea why it works :( Tags are not updated immediately, a dummy buffer change is required to update. Damn.

dungpa:
I tested it locally. It seems to work but it is laggy.

@duckmatt Which lint rules really use type-checked results? I think of using parse results only for linting on file changes. It is currently very slow to lint with type-checked results.

duckmatt:
I think it's probably not worthwhile using the type checked results if they're slowing things down, they're used to double check on a few rules, and without the checks the linter will not warn so you may get a couple of warnings not occurring on edge cases e.g. the naming of a literal when the attribute was a type alias of the literalattribute type. A couple of months back I changed the MSBuild task to not do type checking by default

duckmatt:
The rules it's specifically used for is:

* Check value assigned to binding is not mutable in useless binding rule (will fallback to no warning)
* Check if binding a literal in naming convention rules (will fallback to checking the attribute identifier is `Literal` or `LiteralAttribute`)
* Check if pattern is a union case with no values in naming convention rules (will always issue a warning without type checking)
* Check hints involving the equality operator are not matched against constructor property initialisers or named arguments (will fallback to no warning)

vasily-kirichenko:
It does not update tags after the source was changed:

vasily-kirichenko:
@dungpa I don't think it's laggy because it uses type check results. Almost all other features use them, so what's difference? Or you like linter to work instantly?

vasily-kirichenko:
The good thing in this last bug is that if we fix it, the wrong colorization (after buffer changed) will be fixed, too :)

dungpa:
Yeah, it would be great to fix that bug as part of #1036. 

If not, we could fall back to use untyped ASTs only. We lint a single file at a time so it would be nice to have quick feedback upon changes.

duckmatt:
I've started having a look at UI for the settings, and found they didn't quite match the format of a property grid because the analysers have settings as well as their child rules. I had a look at emulating the settings UI in the [StyleCop extension](http://blogs.msdn.com/cfs-filesystemfile.ashx/__key/communityserver-blogs-components-weblogfiles/00-00-01-37-38-metablogapi/4011.StyleCopProjectSettings1_5F00_thumb_5F00_4FEB89BC.jpg). What do you think?

image

vasily-kirichenko:
Looks great for me.

duckmatt:
Okay great. I've explored it a bit further and now have it so that the configuration file to change the rules for can be selected (with the default selection being the config file in the first common base directory of all the projects). 

image

I've added a tab so that the config files that have been loaded for the projects in the solution can be viewed

image

I'll start working on updating the lint configuration API to make this UI functional

dungpa:
Wow, look great.

dungpa:
@vasily-kirichenko Sent you another PR. We can enable using typed ASTs once https://github.com/fsprojects/VisualFSharpPowerTools/issues/1036 is fixed.

vasily-kirichenko:
Works OK for me, fast and no bugs. 
I think we can merge it as is and add @duckmatt 's settings dialog later.

dungpa:
We merge this PR once the minimal support for F# 4.0 is released (see https://github.com/fsprojects/VisualFSharpPowerTools/pull/1047).

vasily-kirichenko:
After updating all packages AppVeyor fails to build the solution:
```
1) Building C:\projects\visualfsharppowertools\FSharpVSPowerTools.sln failed with exitcode 1.
  2) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5): CreatePkgDef : error : FileNotFoundException: Could not load file or assembly 'FSharp.Core, Version=4.3.0.0, Culture=neutral
, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
  3) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.ModuleHandle.ResolveType(RuntimeModule module, Int32 typeToken, IntPtr* typeInstArgs, Int32 typeInstCount, IntP
tr* methodInstArgs, Int32 methodInstCount, ObjectHandleOnStack type)
  4) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.ModuleHandle.ResolveTypeHandleInternal(RuntimeModule module, Int32 typeToken, RuntimeTypeHandle[] typeInstantia
tionContext, RuntimeTypeHandle[] methodInstantiationContext)
  5) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.Reflection.RuntimeModule.ResolveType(Int32 metadataToken, Type[] genericTypeArguments, Type[] genericMethodArgu
ments)
  6) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.Reflection.CustomAttribute.FilterCustomAttributeRecord(CustomAttributeRecord caRecord, MetadataImport scope, As
sembly& lastAptcaOkAssembly, RuntimeModule decoratedModule, MetadataToken decoratedToken, RuntimeType attributeFilterType, Boolean mustBeInheritable, Object[] attributes, IList derivedAttributes, RuntimeType& attributeType, IRunt
imeMethodInfo& ctor, Boolean& ctorHasParameters, Boolean& isVarArg)
  7) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.Reflection.CustomAttribute.GetCustomAttributes(RuntimeModule decoratedModule, Int32 decoratedMetadataToken, Int
32 pcaCount, RuntimeType attributeFilterType, Boolean mustBeInheritable, IList derivedAttributes, Boolean isDecoratedTargetSecurityTransparent)
  8) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.Reflection.CustomAttribute.GetCustomAttributes(RuntimeType type, RuntimeType caType, Boolean inherit)
  9) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at System.RuntimeType.GetCustomAttributes(Boolean inherit)
 10) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at Microsoft.VisualStudio.Tools.CreatePkgDef.ProcessAssembly(String fileName, Hive hive, PkgDefContext context, Boolean r
egister, RegistrationMode mode) in f:\dd\vssdk\VSIntegration\Tools\src\CreatePkgDef\CreatePkgDef.cs:line 270
 11) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at Microsoft.VisualStudio.Tools.CreatePkgDef.DoCreatePkgDef(InputArguments inputArguments) in f:\dd\vssdk\VSIntegration\T
ools\src\CreatePkgDef\CreatePkgDef.cs:line 164
 12) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):    at Microsoft.VisualStudio.Tools.CreatePkgDef.Main(String[] arguments) in f:\dd\vssdk\VSIntegration\Tools\src\CreatePkgDef
\CreatePkgDef.cs:line 85
 13) : C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\VSSDK\Microsoft.VsSDK.targets(760,5):
```

dungpa:
I once fixed it at https://github.com/vasily-kirichenko/FSharpVSPowerTools/commit/67a23acbadbc9da18c322fa071bcc8795e14d347.

vasily-kirichenko:
Thanks!

vasily-kirichenko:
Please don't touch `GeneralOptionsControl` anymore until this PR is merged. It's impossible to merge WinForms conflicts.

dungpa:
OK.

dungpa:
Great jobs. This PR can now be merged.

PR #1038, opened 2015-07-13
NavigateToSource now supports bitbucket urls

TWith2Sugars:
This should allow support for bitbucket urls when navigating to source. Bitbucket doesn't have a "raw" subdomain and the link anchor is different to github.

Haven't managed to get this installed locally but I'll give it a try when the appveyor ci builds this branch (hopefully).

### DIFF
--- a/src/FSharpVSPowerTools.Core/Utils.fs
+++ b/src/FSharpVSPowerTools.Core/Utils.fs
@@ -376,6 +376,20 @@ module String =
     let split options (separator: string[]) (value: string) = 
         match value with null -> null | x -> x.Split(separator, options)
 
+    let (|StartsWith|_|) value x =
+        if String.IsNullOrWhiteSpace(x) then
+            None
+        elif x.StartsWith(value) then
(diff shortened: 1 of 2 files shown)


TWith2Sugars:
So after a few false starts it works nicely with bitbucket.

dungpa:
Could you define `(|StartsWith|_|)` active patterns somewhere in `FSharpVSPowerTools.Core` project and use it here? Then the cases of `githubUserContent` and `githubRaw` could be merged into an or pattern e.g.

```fsharp
match url with
| StartsWith "https://raw.githubusercontent.com"
| StartsWith "https://github.com" -> ...```

TWith2Sugars:
Like so?

I've also included support for [codebasehq](https://www.codebasehq.com/) urls as well.

dungpa:
That was it. Great jobs.

forki:
Kudos

ctaggart:
Just now seeing this. Excellent work!

Issue #1048, opened 2015-07-21
Error in VS2015 RTM


simra:
After installing F# power tools in VS2015 I see this error on launch:

> The 'PowerToolsCommandsPackage' package did not load correctly.

> The problem may have been caused by a configuration change or by the installation of another extension. You can get more information by examining the file '...\AppData\Roaming\Microsoft\VisualStudio\14.0\ActivityLog.xml'.

> Restarting Visual Studio could help resolve this issue.

> Continue to show this error message?

Examining the ActivityLog, I see this error message:
>     <errorinfo>The type initializer for 'ForegroundThreadGuard' threw an exception.</errorinfo>

simra:
i probably put the cart before the horse because I installed this first, before opening any F# project that would trigger installation of the Visual F# components.  After installing Visual F#, opening a .fs file caused VS to crash.  I disabled the power tools and can open .fs files safely.  Reinstalled power tools and now everything works fine. I'll close, feel free to keep around as a low-pri bug.

smoothdeveloper:
Considering that F# language has probably the lowest footprint of language support compared to other .net languages or c++, it would make sense to have it installed by default.

Disabling features based on telemetry makes sense when the features weight more than a couple hundred megabytes, but I believe F# is still bellow this threshold.

Also it seems lazy installation works for projects but doesn't work on opening individual .fs / .fsi / .fsx files, nor when installing plugins relying on the language service being installed.

Since VS2010, opening an .fsx on a default VS install would provide intellisense, full language support and interpreter interaction, that was a selling point for using F# for exploratory coding, that seems like a step backward to get people to use the language.

PR #1056, opened 2015-07-24
Set VSIX dependency on Visual F# Tools

dungpa:
This is an attempt to prevent #1048.

Drop support for VS 2012 since it's not possible to set dependency on VS 2012.

I will create a `vs2012` branch and have an archived vsix on AppVeyor for anyone who stucks with VS2012.


vasily-kirichenko:
I cannot test whether #1048 is fixed.

Issue #863, opened 2014-11-16
Running out of space to add disabling options for new features


dungpa:
The "General Options" dialog needs to be redesigned to accommodate more options. Does anyone have any ideas to address this issue?

dungpa:
Marked as up-for-grabs. Hoping for a UI guru to appear and solve this :-).

dungpa:
@xperiandri Would you mind to help with this :-)? 

The idea is to replace all checkboxes by a property grid which is scrollable when adding more options. It works similar as Productivity Power Tools' options dialog.

xperiandri:
Would you propose to use toggle switches like they do?

Although it is somewhat harder to read property pages in general this is a good idea. If you group options into categories it is easier to find. I looked at Web Essentials options and they look pretty well

Another option is the way it is done in Debugging.

dungpa:
A property grid with True/False would be enough I think. Productivity Power Tools' toggle switches are fancier but they can be confusing sometimes.

I agree that grouping options is very natural. We have similar approaches for formatting options at http://fsprojects.github.io/VisualFSharpPowerTools/codeformatting.html.

xperiandri:
So, would you like me to do that?
Move from checkboxes to property grid?

I think it is worth to do a release with my layout improvement before I do that so we can compare usability.

dungpa:
Yes, please. You can choose a way that you find the most suitable. The main goal is to be able to accommodate more options.

We bump versions very often. You can download the latest version at https://ci.appveyor.com/project/dungpa/visualfsharppowertools/branch/master/artifacts.

xperiandri:
But the latest release does not have my UI changes.

dungpa:
Your UI changes have been released with v2.0.0. It would be great if you help resolve this issue. It blocks us from adding any new feature to VFPT :-).

xperiandri:
I see
Cool
Next week

xperiandri:
So now it might look like you asked

xperiandri:
What is Linter? LinterEnabled property?
Category
DisplayName
Description
DefaultValue

dungpa:
```csharp
[Category("Editor enhancements")]
[DisplayName("Linter")]
[Description("Enable or disable FSharpLint integration.")]
[DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
[DefaultValue(true)]
public bool LinterEnabled { get; set; }
```

[...]

PR #1062, opened 2015-07-28
Changed General options page from custom to property grid layout

xperiandri:
Changed General options page from custom to property grid layout according to issue #863


dungpa:
@xperiandri The PR is non-mergeable. Could you rebase to current master?

vasily-kirichenko:
It does not compile in VS 2013

image

VS 2015 is buggy as hell. We should not use any F# 4 / C# 6 features yet. We may consider this after Update 1 releases.

xperiandri:
Does settings application work only after closing, and reopening code file?

dungpa:
Most settings work after reopening files. Task List comments, Folder organization and Generate references for FSI settings require restarting Visual Studio to take effects.

xperiandri:
In Visual Studio 2015 with Update 5 nameof works
But I will change that

@vasily-kirichenko, I agree that 2015 is very buggy but benefits worse.
It crushes for dozens times a day. I don't used to use so much foul language as after RTM

xperiandri:
Why GeneralOptionsPage opens on Visual Studio lunch but not when F# Power Tools in Options dialog is selected?

xperiandri:
@dungpa, try to merge now
Everything have to work.
By the way, how to run tests?

xperiandri:
@vasily-kirichenko, looks like I have managed to fix VS2015 and now it works stable

dungpa:
Check out the Contributing guide https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/CONTRIBUTING.md

In short, you can run 'build.cmd' from command line to run tests.

xperiandri:
Will not work running them within Test Explore in Visual Studio?

dungpa:
Yes, it does work if NUnit Test Adapter is installed https://visualstudiogallery.msdn.microsoft.com/6ab922d0-21c0-4f06-ab5f-4ecd1fe7175d

dungpa:
Quick comments:

 - Please don't rename `FSharpVSPowerTools.sln`. It breaks our build on AppVeyor CI https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/2.1.0-a1507292342
 - Please don't rearrange `FSharpVSPowerTools.csproj`; it's not the purpose of this PR.

Initially we didn't use the property grid supported in VS for General Options dialog because it has rigid structures. For example, we want to preseve the 'You must restart Visual Studio for the changes to take effects' label. How do we do it with this property grid?

From current version of `GeneralOptionsControl`, I think it would be even simpler to use `PropertyGrid` controls from ToolBox and bind data ourselves.

xperiandri:
I can't see any purpose to do it so complex.
We have only 1 option that require restart as far as I know.
I implemented logic to gray it out if Visual Studio is not running with admin privileges. There is a *** sign on that property and a description.
If user not an idiot and he didn't read a description of that option he must guess that restart required after not appearing or disappearing a Navigation Bar.

I insist that this label about required restart used to be completely redundant unless there are other options that require restart.

xperiandri:
@dungpa, are you sure that you will not ___rearrange `FSharpVSPowerTools.csproj`___?
Because it will take additional hour to bring it back.

xperiandri:
We can change
```F#
let solutionFile  = "FSharpVSPowerTools"
```
to
```F#
let solutionFile  = "FSharpVSPowerTools.v12"
```
in `build.fsx`
It is not as confusing for new contributors as existence of some sln files with suffixes and one without

dungpa:
This PR should have been split into three different ones.

The first one deals with formatting changes. We could easily review formatting changes in isolation.

The second PR rearranges `FSharpVSPowerTools.csproj`. Your folder grouping is inconsistent and confusing. I suggest:
 - We have `PowerToolsCommandPackage.cs` is an entry point in the top level folder. 
 - A `Settings` folders for all `*OptionsControl.cs` and `*OptionsPage.cs` files.
 - A `Features` folders for the remaining facades to the available features

The third PR changes GeneralOptions dialog to use a property grid.

> I insist that this label about required restart used to be completely redundant unless there are other options that require restart.

The three options that requires restarting Visual Studio are:

 - Folder organization
 - Task List comments
 - Generate references for FSI

If we count 'Navigation Bar', it's going to be **four** features doing so.

It's not only about keeping the label **as is**. The current dialog allows flexible changes to the UI. My opposition is that we can move it to a property grid in a simple way without losing flexibility.

xperiandri:
As you wish

new general options page

xperiandri:
> •We have  PowerToolsCommandPackage.cs  is an entry point in the top level folder. 
> •A  Settings  folders for all  *OptionsControl.cs  and  *OptionsPage.cs  files.
> •A  Features  folders for the remaining facades to the available features
What is the benefit?

I thought reorganization to increase seeking speed.
If I want to change something about General page I go to General folder
If I want to change something about Formatting page I go to formatting
etc

What is your idea?

dungpa:
> If I want to change something about General page I go to General folder
> If I want to change something about Formatting page I go to formatting
etc

Your idea is right. But you initially push https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools/SyntaxConstructClassifierProvider.cs to `General` folder, which is not quite consistent. These facade files for specific features should be a separate folder.

xperiandri:
OK, it was a mistake

vasily-kirichenko:
How this property grid is better than the current check boxes? It's very difficult to see at a first glance what features are enabled in that grid. Words "True" and "False" vs [ ] and [x].

xperiandri:
I can agree with your point
Other extensions use their own windows:
- CodeMaid - modal windows
- vsCommands - tool window
But I don't think that this is a good idea. I prefer to be along with standard ways of doing things.

As I said earlier the other option is to use a control like in Debugging. But it is quite ugly.

xperiandri:
@dungpa, do you have any remarks about code in current pull request?

dungpa:
The code is fine.

Some problems:

1. `DefaultValue` attributes seem problematic. XML comments generation is always persisted as `False` in the next VS session no matter I change it. Perhaps we should assign values for properties in constructors as before?

2. Typo: 'Enables or disables' --> 'Enable/disable'

3. Typo: 'Refactorings' --> 'Refactoring'

4. Typo: 'requires restart' --> 'Require restart'

5.Typo: 'requires admin' --> 'Require admin'

xperiandri:
I thought that it is implied option at the beginning of the sentence that is why I spelt
"Option requires restart"
"Option requires admin privileges"
Isn't it correct?

I will check about XML comments.
Default attribute just allows to specify that true or false won't be bold.

xperiandri:
About default values you are right.
I will bring that back.

Could you explain
If I type /// and press space on top of function that has parameters it must added necessary XML comments to describe parameters, mustn't it?

xperiandri:
Fixed properties initialization

dungpa:
> I thought that it is implied option at the beginning of the sentence that is why I spelt it from the third person in singular

A sentence starting with a verb assumes that the subject is neutral. You can check other options in Visual Studio to be sure.

> Could you explain
> If I type /// and press space on top of function that has parameters it must added necessary XML comments to describe parameters, mustn't it?

In F#, one can use `/// comments` as the short-form comments. To activate the feature, type `///<` http://fsprojects.github.io/VisualFSharpPowerTools/xmldoc.html

Btw, you can push more commits. No need to force push every time.

vasily-kirichenko:
I still think this PR makes the option page worse. Why this PR at all?

dungpa:
@vasily-kirichenko Do you have a better way to find spaces for new options? 

We use a fairly standard way (similar to Productivity Power Tools) here. It's not too bad given that we have very constrained spaces in the option dialog.

vasily-kirichenko:
What new options? I don't have any issues with options page at the moment. When (if) we add few new features (in a year or two), we could consider reorganizing the page. Doing it now is a waste of time, especially considering the result.

xperiandri:
> Btw, you can push more commits. No need to force push every time.

Why to keep irrelevant commits?

> A sentence starting with a verb assumes that the subject is neutral. You can check other options in Visual Studio to be sure.

No, have a look at Windows Forms Designer options page
> Toggles ...
> Controls ...

VC++ Project Settings: Show Environment In Log
> Specifies whether ...

Meaning that "This option" text is omitted at the beginning of the sentence.

dungpa:
> Meaning that "This option" text is omitted at the beginning of the sentence.

Fair point. I would say VS is not entirely consistent in this regard. Most of their options don't use that form though. If you don't want to change this, it's fine.

> Why to keep irrelevant commits?

You update the PR based on feedback. If you added more comments, we can see the delta changes to understand the progress. As it is, we have to look through all the code again.

dungpa:
@vasily-kirichenko We ran out of spaces long ago. To be honest, not everyone has high DPI screens or large monitors. The options look so dense and they lack spaces on my poor laptop. I feel uncomfortable with squeezing spaces of the check boxes everytime we add a new feature.

If we should do this sooner or later. Why don't we do it now?

xperiandri:
> If you added more comments, we could see the delta changes to understand the progress. As it is, we have to look through all the code again.

Agree.
My intent was to revert everything back in case of splitting that big pull request. And to keep commit consistent (as again I reverted some part of code).
So I decided not to do a commit which deletes something and then reverts it back

dungpa:
With your changes, `XmlDocEnabled` still doesn't work for me. It has been persisted correctly to registry at `HKCU\Software\Microsoft\VisualStudio\12.0\DialogPage\FSharpVSPowerTools.GeneralOptionsPage\XMLDocEnabled`, however, its value isn't displayed correctly on the property grid. 

Somehow, the dynamic way of passing attributes to `GeneralOptionsControl` screwed up reading registry values.

@xperiandri Did you check whether it works locally on your machine?

xperiandri:
Yes
I did settings reset:
Deleted 14.0Exp folders from Local and Roaming
And performed settings import, C# profile.
After this everything worked well

dungpa:
I don't think we should go down this route anymore. 

It means a subtle bug for users. If they don't reset settings by themselves, the real stored settings and displayed values on the property grid are out of sync.

I'm terribly sorry. I didn't know that it is this tricky. 

At best, we can add a scrollable Panel to have more spaces for new options.

xperiandri:
1. The is already a scrollable control with adaptive layout.
2. Why do you care so much?

We need to test in clean environment.
At first install current version and change some settings.
At second update to this version with my changes and check.

Even though it will require to reset settings it is not so big deal. We just need to specify it in extensions description on Visual Studio Gallery and set to open this page after update.

dungpa:
I tested it by bumping version and installed this version on top of v2.0.0. It has weird behaviours if users update from some old VFPT version.

This is a small improvement on top of something that has been working OK. The downside is breakage to current users. I don't think the benefit outweighs the cost of adding the improvement.

> Why do you care so much?

If I didn't care, I wouldn't put so much effort into this project.

Again I'm really sorry that you have made a lot of effort in order that we figured out that it didn't work as anticipated. 

We still have a lot of pending issues where we hope for community's contributions, especially up-for-grabs ones https://github.com/fsprojects/VisualFSharpPowerTools/labels/up-for-grabs. We would love to have your contributions on those.

xperiandri:
It is not only about effort, it is about usability.
> If I didn't care, I wouldn't put so much effort into this project.

Any way there is no alternative to this project. You can do whatever you think is better.

> Again I'm really sorry that you have made a lot of effort in order that we figured out that it didn't work as anticipated. 

I will double check later.

>  The downside is breakage to current users.

So what? It is not a big deal. Adding "Known issues" paragraph to release notes will fix the problem.

Issue #444, opened 2014-05-24
Almost nothing works for a code using Fleece 


vasily-kirichenko:
```fsharp
module FleeceTest.Test

#r @"..\packages\System.Json.4.0.20126.16343\lib\net40\System.Json.dll"
#r @"..\packages\ReadOnlyCollectionInterfaces.1.0.0\lib\NET45\ReadOnlyCollectionsInterfaces.dll"
#r @"..\packages\Fleece.0.2.0\lib\NET40\Fleece.dll"

open Fleece

type T() =
   static member FromJSON (_: int) = function
       | JObject o -> printfn "%A" o 
       | _ -> ()
```
FCS does not return anything for line `| JObject o -> printfn "%A" o`.

vasily-kirichenko:
@dungpa no, it has not been fixed. VS 2013:

![image](https://cloud.githubusercontent.com/assets/873919/9025347/e2857de6-3909-11e5-857e-67de16a79339.png)

No coloring, no current symbol highlighting and so on.

dungpa:
This version works (notice changine 'NET45' to 'Net40-client'):

```fsharp
module FleeceTest.Test

#r @"..\packages\System.Json.4.0.20126.16343\lib\net40\System.Json.dll"
#r @"..\packages\ReadOnlyCollectionInterfaces.1.0.0\lib\Net40-client\ReadOnlyCollectionsInterfaces.dll"
#r @"..\packages\Fleece.0.2.0\lib\NET40\Fleece.dll"

open Fleece

type T() =
   static member FromJSON (_: int) = function
       | JObject o -> printfn "%A" o 
       | _ -> ()
```
I guess the reason is that FSI still uses .NET 4.0 assemblies for resolution.

Here is the resolution with project files. It works fine since all assemblies are resolved in .NET 4.5:

> GetProjectCheckerOptions: ProjectFileName: C:\Users\abc\Documents\Visual Studio 2013\Projects\TutorialPerf1\TutorialPerf1.fsproj, ProjectFileNames: [|"C:\Users\abc\AppData\Local\Temp\.NETFramework,Version=v4.5.AssemblyAttributes.fs";
  "C:\Users\abc\Documents\Visual Studio 2013\Projects\TutorialPerf1\Library1.fs"|], FSharpProjectOptions: [|"-o:obj\Debug\TutorialPerf1.dll"; "-g"; "--debug:full"; "--noframework";
[...]

PR #1065, opened 2015-08-01
Migrate to new FCS compatible with F# 4.0

dungpa:
The idea is that we use FSharp.Core 4.4.0.0 for VS 2015 and crack compiler options to use FSharp.Core 4.3.1.0 for VS 2013 and FSharp.Core 4.3.0.0 for VS 2012.


dungpa:
As a side effect, it may fix https://github.com/fsprojects/VisualFSharpPowerTools/issues/444. Could you please check?

vasily-kirichenko:
Works OK on a machine with VS 2013 and VS 2015 installed (in both VSs).

dungpa:
I think it's OK to merge this. Will investigate #444 further.

PR #1070, opened 2015-08-07
Skip object identifiers when checking for unused symbols

dungpa:
UserVoice request: http://vfpt.uservoice.com/forums/247560-general/suggestions/8793808-do-not-mark-orange-in-the-right-bar-unused-self-re

Graying out object identifiers could be annoying. Since it has little practical value, we skip it while checking for unused references.

### DIFF
--- a/src/FSharpVSPowerTools.Core/TypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/TypedAstUtils.fs
@@ -302,6 +302,8 @@ module UnusedDeclarations =
             | Entity ((Record | UnionType | Interface | FSharpModule), _, _) -> None
             // FCS returns inconsistent results for override members; we're going to skip these symbols.
             | MemberFunctionOrValue func when func.IsOverrideOrExplicitInterfaceImplementation -> None
+            // Ignore object identifiers i.e. 'this' since graying them out could be annoying. 
+            | MemberFunctionOrValue func when func.IsMemberThisValue -> None
             // Usage of DU case parameters does not give any meaningful feedback; we never gray them out.
             | Parameter -> None
             | _ ->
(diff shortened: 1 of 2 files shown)


vasily-kirichenko:
I'm afraid I disagree. What's more, that User Voice has only 4 votes. I like how it works now. Maybe we should ask people?

dungpa:
Sure, please ask around on Twitter and Slack.

vasily-kirichenko:
@cloudroutine wrote on Slack:

>  I'd prefer another toggle like "Gray out unused member identifiers" and a toggle for whether unused items show up in the scroll bar

I agree, we should add two options:

* Syntax coloring
 * Gray out unused opens
 * Gray out unused declarations
   * Gray out unused self identifiers
 * Show unused symbols on scroll bar 

BTW, what do you think about using "Highlight" instead of "Gray out"?

cloudRoutine:
Highlight is more accurate since this is in the context of syntax highlighting and because the color for unused items doesn't have to be grey.

dungpa:
Gray out unused self identifiers: the option is too fine-grained to support a very specific use case. I wouldn't want to go that way. 

Show unused symbols on scroll bar: It might be useful, but it doesn't solve the problem here.

> What do you think about using "Highlight" instead of "Gray out"?

Yes, it's better with 'Highlight' as @cloudRoutine said.

vasily-kirichenko:
If you don't like the idea to add a setting for "Gray out unused self identifiers" and not all people like removing the feature (including me), I suggest to drop this PR.

ashtonkj:
Please do make it optional. I consider the "this" identifier to be used as soon as a member is declared using the member this.membername syntax. I don't like being told that it isn't being used when it is required by the syntax (it feels like returning a false positive).

vasily-kirichenko:
`this` is completely redundant. It's just a variable, and if you don't need it, you should claim about it explicitly, like `_self` or `__`. I think putting `this` everywhere is just old C# habit.

dungpa:
This tweet convinced me https://twitter.com/CarstenK_Dev/status/631013723700903936. 

Happy to close this and forget it forever.

PR #1074, opened 2015-08-08
Hide the lightbulb when there is no current word

### LINKED ISSUE #1073: Lightbulb still appears when the cursor is out of current word
The lightbulb is still on after the cursor moves to spaces.

dungpa:
Close #1073.


vasily-kirichenko:
Great, thanks!

Issue #1037, opened 2015-07-08
Slows down loading huge solution file


unuunc:
Hi ,

Can we disable the extension if project type is not of FSharp language projects. Observed that this extension slows down loading of my solutions file (very large, >100 projects ) even if that solution does not contain any Fsharp code.

I tried by enabling and disabling and it seems this extension is the root cause.

Thanks

dungpa:
What features do you enable in the PowerTools setting? 

Only two features that do something while loading solutions are Task List comments and Folder Organization. Could you identify which feature causes the issue? Thanks.

unuunc:
Tried your suggestion and it is faster after disabling these features. Still we can think of enabling extension only in case when project has FS code.

vasily-kirichenko:
How much it's slower? 10x? 1.0001x?

dungpa:
The extension doesn't do anything after knowing that the solution doesn't contain F# projects. However, checking  project types in solutions are expensive.

xperiandri:
Maybe it is because of usage of old API? Could using EnvDTE90, 100 help?

vasily-kirichenko:
@unuunc can you provide measurements? How longer your solution is loaded having VFPT enabled? If you cannot, I'll close this issue.

0x53A:
I have the same issue.
We have a solution with about 60 C# Projects, two C++ Projects and two F# Projects.

Without VFSPT, it loads in a few seconds (<10sec).
With VFSPT, it takes in the range of minutes.
(I don't have access to it at the moment, so can't measure it exactly).

What is annoying, is that changing the GIT branch triggers a (partial) solution reload, and waiting minutes after changing a branch takes a lot of the fun out of git.

I haven't had time for a detailed analysis, but one thing I noticed was that it spammed
[Language Service] Clearing FCS caches.
in the debug output.

vasily-kirichenko:
Thanks a lot for the investigation. I suspect those three full synchronous GC collects are the problem. Will look at it.

vasily-kirichenko:
@dsyme can you recommend a faster way to clean FCS caches (compared to `ClearLanguageServiceRootCachesAndCollectAndFinalizeAllTransients`)?

vasily-kirichenko:
@0x53A Please try this VSIX https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/2.1.0-a1508090818/artifacts

[...]

Issue #474, opened 2014-06-07
High memory consumption


vasily-kirichenko:
Open our solution, build it, switch to another branch, build again, switch to one other branch, build one more time. All this in 5 minutes interval. The result is as following: 

![image](https://cloud.githubusercontent.com/assets/873919/3208203/d6747900-ee21-11e3-86f5-8412f940fe35.png)

This is all _without_ executing Find all references / Rename refactoring at all. 
As a result, everything is greatly slowed down, VS periodically freezes for some seconds, then redraw itself and so on. 

When I profiled it with PerfView I didn't see any leakage in our code, all these hundreds of megabytes were occupied by `byte[]` arrays related to "IL cache" inside FCS (don't know what it is though). So, I suspect FCS caching does not behave nicely for our scenarios.

As our solution is small, I think we have a big problem here.

PR #1076, opened 2015-08-09
Clearing FCS caches optimization

vasily-kirichenko:
WIP


dungpa:
Looking at old conversations (e.g. https://github.com/fsprojects/VisualFSharpPowerTools/issues/474), `ClearLanguageServiceRootCachesAndCollectAndFinalizeAllTransients` was added when we chased memory leaks.

Now we know that it's mostly because of type providers and VFT itself, do you think there is any problem with merging this?

vasily-kirichenko:
As it's fixed https://github.com/fsprojects/VisualFSharpPowerTools/issues/1037, I think we should merge this. I'll use this version at work and will see whether it's introduced any leaks.

dungpa:
:shipit:

PR #1092, opened 2015-08-22
Fix unit test execution under VS 2015

dungpa:
- Use FSharp.Core from NuGet package
- Add binding redirects for VS assemblies


vasily-kirichenko:
The tests are loaded into NUnit.exe and passed. Wow! :)

Issue #1094, opened 2015-08-23
Complete FSharpLint integration


dungpa:
There are only two remaining things to finish:

 - Add UI for changing Lint settings
 - Add docs for FSharpLint integration

dungpa:
@duckmatt When you have a prototype UI for changing FSharpLint settings, please send a PR. We could help a bit in order to complete the feature.

duckmatt:
Okay great, I've got a fully functioning UI for the options now, looking at hooking it up to the new configuration management I've been adding to FSharpLint. I'll get a work in progress pull request together

dungpa:
It has been done in v2.1.0. Thanks Matt for the great work.

duckmatt:
No problem, thank you for including it

PR #1097, opened 2015-08-25
[WIP] Integration of FSharpLint

duckmatt:
This pull request is to finish up the intgeration of FSharpLint by providing a configuration pane in the power tools options. #1094 

The work here so far is mostly to display an existing (default) configuration file in the options. Going forwards I'm looking to integrate it with the configuration management functionality I've been adding to FSharpLint so that any updates made by the user are saved back to disk, and then cleaned up into a state where it can be merged.


dungpa:
Great jobs. I think it needs to be hooked with config write API then we're done.

Time for some screenshots:

lint_settings

lint_rules

src/FSharpVSPowerTools.Logic/LintTagger.fs
```diff
@@ -31,9 +31,28 @@ type LintTagger(textDocument: ITextDocument,
             let source = buffer.CurrentSnapshot.GetText()
             let! parseFileResults = vsLanguageService.ParseFileInProject (doc.FullName, source, project) |> liftAsync
             let! ast = parseFileResults.ParseTree
+
+            let config = 
+                match ConfigurationManagement.loadConfigurationForProject project.ProjectFileName with
+                | ConfigurationManagement.ConfigurationResult.Success(config) -> Some(config)
```
dungpa: Is this possible to `open ConfigurationManagement`?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,218 @@
...
+        with get() = value.Value
+        and set(v) = value.Value <- v
+
+type HintsViewModel(config:Configuration) as this =
+    inherit ViewModelBase()
+
+    let validateHint hint =
+        [
+            if hint <> "" then
+                match FParsec.CharParsers.run FSharpLint.Framework.HintParser.phint hint with
```
dungpa: Can we `open FParsec` and `open FSharpLint.Framework`?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,218 @@
...
+                match FParsec.CharParsers.run FSharpLint.Framework.HintParser.phint this.NewHint with
+                | FParsec.CharParsers.Success(_) -> 
+                    hints.Add(this.NewHint)
+                    this.NewHint <- ""
+                    this.SelectedHintIndex <- hints.Count - 1
+                | FParsec.CharParsers.Failure(_) -> ())
+
+    member this.RemoveHintCommand = 
+        this.Factory.CommandSyncParam(fun (selectedItem:obj) ->
+            if selectedItem :? string then
```
dungpa: If we use `match selectItem with :? string as s` then `s` is string and we don't need downcasting (`:?>`) anymore?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,218 @@
...
+                | IgnoreBlankLines(value) -> 
+                    yield BoolViewModel("IgnoreBlankLines", value) :> obj
+                | Access(value) -> 
+                    yield AccessViewModel("Access", value) :> obj
+                | Hints(_)
+                | Enabled(_) -> ()
+        ]
+
+    let isRuleEnabled (settings:Map<string, Setting>) =
+        if settings.ContainsKey "Enabled" then
```
dungpa: Use `Map.tryFind`?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,218 @@
...
+    member this.AddIgnoreFileCommand = 
+        this.Factory.CommandSync(fun _ -> 
+            if String.IsNullOrEmpty this.NewIgnoreFile |> not then
+                ignoreFiles.Add(this.NewIgnoreFile)
+                this.NewIgnoreFile <- ""
+                this.SelectedIgnoreFileIndex <- ignoreFiles.Count - 1)
+
+    member this.RemoveIgnoreFileCommand = 
+        this.Factory.CommandSyncParam(fun (selectedItem:obj) ->
+            if selectedItem :? string then
```
dungpa: Please use pattern matching and `as` pattern as above.

src/FSharpVSPowerTools.Logic/Linting/LintOptionsPage.fs
```diff
@@ -0,0 +1,39 @@
...
+open System.Windows
+open System.ComponentModel.Composition
+open Microsoft.VisualStudio.OLE.Interop
+open System.Collections.Generic
+
+[<Guid("f0bb4785-e75a-485f-86e8-e382dd5934a4")>]
+type LintOptionsPage() =
+    inherit UIElementDialogPage()
+
+    let getProjectPaths () =
```
dungpa: This function is currently unused? I guess it's needed when we load and save configs.

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
duckmatt: Thanks, turned out `CommandSyncParam` was generic and `:obj` could be removed

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
duckmatt: Thanks, updated to use `tryFind`

src/FSharpVSPowerTools.Logic/Linting/LintOptionsPage.fs
duckmatt: Yeah, this will be used to help load the config files from disk

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,215 @@
...
+open FParsec
+open FSharp.ViewModule
+
+type BoolViewModel(name, isChecked) as this =
+    inherit ViewModelBase()
+
+    let isChecked = this.Factory.Backing(<@ this.IsChecked @>, isChecked)
+
+    member this.Name
+        with get() = name
```
vasily-kirichenko: `member __.Name = name`?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,215 @@
...
+                | Access(value) -> 
+                    yield AccessViewModel("Access", value) :> obj
+                | Hints(_)
+                | Enabled(_) -> ()
+        ]
+
+    let isRuleEnabled (settings:Map<string, Setting>) =
+        match settings.TryFind "Enabled" with
+        | Some(Enabled(e)) -> e
+        | _ -> false
```
vasily-kirichenko: `settings.TryFind "Enabled" = Some (Enabled true)`?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,215 @@
...
+    let selectedRule = this.Factory.Backing(<@ this.SelectedRule @>, None)
+    let newIgnoreFile = this.Factory.Backing(<@ this.NewIgnoreFile @>, "")
+
+    let ignoreFiles =
+        match config.IgnoreFiles with
+        | Some(x) -> 
+            x.Content.Split([|Environment.NewLine|], StringSplitOptions.RemoveEmptyEntries)
+                |> Seq.map (fun x -> x.Trim())
+                |> (fun x -> ObservableCollection<_>(x))
+        | None -> ObservableCollection<string>()
```
vasily-kirichenko: We have some useful functions over here https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Core/Utils.fs#L349-L391
```fsharp
let ignoreFiles =
    ObservableCollection(
        match config.IgnoreFiles with
        | Some x -> 
            x.Content 
            |> String.split StringSplitOptions.RemoveEmptyEntries [| Environment.NewLine |]
            |> Array.map String.trim
        | None -> [||])
```

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
```diff
@@ -0,0 +1,215 @@
...
+    let selectedIgnoreFileIndex = this.Factory.Backing(<@ this.SelectedIgnoreFileIndex @>, 0)
+
+    let rules = SetupViewModels.ruleViewModelsFromConfig config
+
+    member this.SelectedRule 
+        with get() = selectedRule.Value
+        and set (value) = selectedRule.Value <- value
+    
+    member this.Files
+        with get() = files
```
vasily-kirichenko: `... = files` and so on for all properties.

src/FSharpVSPowerTools.Logic/Linting/RuleViewModel.fs
```diff
@@ -0,0 +1,47 @@
...
+open System.ComponentModel
+open FSharp.ViewModule
+
+type RuleViewModel(name:string, rules:RuleViewModel seq, settings:obj seq) as this =
+    inherit ViewModelBase()
+
+    let isChecked = this.Factory.Backing(<@ this.IsChecked @>, false)
+            
+    member this.Name
+        with get() = name
```
vasily-kirichenko: same.

src/FSharpVSPowerTools.Logic/Linting/RuleViewModel.fs
```diff
@@ -0,0 +1,47 @@
...
+
+    member this.IsChecked
+        with get () = isChecked.Value
+        and set (value) = 
+            isChecked.Value <- value
+
+            match this.ParentRule with
+            | Some(parentRule:RuleViewModel) ->
```
vasily-kirichenko: Why type annotation here?

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
duckmatt: Thanks, I hadn't realised you could shorten get properties to that

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
duckmatt: Thanks, updated to use those functions

src/FSharpVSPowerTools.Logic/Linting/OptionsViewModel.fs
duckmatt: Thanks, that's a nice simplification I hadn't seen

duckmatt:
This is mostly functional now, needs a bit of cleaning up as I've just been trying to get something working (been a bit strapped for time):

* Needs dialog to confirm whether to save changes to current config when switching to another config
* Needs dialog to notify the user when unable to save to a config (file could be locked, no write permission etc)
* Handling of config paths needs to be ensured it's consistent
* Needs testing
* Config updating added to FSharpLint needs further testing

vasily-kirichenko:
image

No hint:

image

duckmatt:
Your hint will match:
    
    ``fun`` ()

I'll update the hint dsl to not parse keywords as identifiers if they're not in backticks to avoid this confusion

Unfortunately the lambda based hints aren't very clever at matching the lambda's parameters at the moment, the hint would be `fun () -> _ ===> fun _ -> _` but currently won't be matched

src/FSharpVSPowerTools.Logic/Linting/FileViewModel.fs
```diff
@@ -0,0 +1,21 @@
...
+    member __.Files = files
+
+    member __.HasConfigFile = hasConfigFile
+
+    member __.GetDirectory() =
+        let directorySeparator = Path.DirectorySeparatorChar.ToString()
+        String.concat directorySeparator path
+
+    member this.GetFilePath() =
+        Path.Combine(this.GetDirectory(), SettingsFileName)
```
vasily-kirichenko: Use `</>` for combining paths.

src/FSharpVSPowerTools.Logic/Linting/LintOptionsPage.fs
```diff
@@ -0,0 +1,87 @@
...
+open System.Diagnostics
+open System.IO
+open System.Runtime.InteropServices
+open System.Windows
+open System.ComponentModel.Composition
+open Microsoft.VisualStudio.OLE.Interop
+open System.Collections.Generic
+open FSharpLint.Framework.Configuration
+open Management
+open LintUtils
```
vasily-kirichenko: Remove unused open declarations.

src/FSharpVSPowerTools.Logic/Linting/LintOptionsPage.fs
```diff
@@ -0,0 +1,87 @@
...
+    new () = new LintOptionsPage(None)
+
+    interface ILintOptions with
+        member this.UpdateDirectories() =
+            loadedConfigs <- updateLoadedConfigs (this.GetDte()) loadedConfigs
+
+        member this.GetConfigurationForDirectory(dir) =
+            getConfigForDirectory loadedConfigs dir
+
+    member private this.GetDte() =
```
vasily-kirichenko: `this` is unused.

src/FSharpVSPowerTools.Logic/Linting/LintOptionsPage.fs
```diff
@@ -0,0 +1,87 @@
...
+        match dte with
+        | Some(dte) -> dte
+        | None -> this.GetService(typeof<EnvDTE.DTE>) :?> EnvDTE.DTE
+            
+    override this.OnApply(_) =
```
vasily-kirichenko: as above.

src/FSharpVSPowerTools.Logic/Linting/LintOptionsPage.fs
```diff
@@ -0,0 +1,87 @@
...
+                        file) |> Some
+                | None -> 
+                    Debug.Assert(false, "No file view model for the initial file found.")
+                    None
+            | None -> 
+                None
+
+        lintOptionsPageControl.Value.DataContext <- LintViewModel(lintOptions, saveViewModel)
+            
+    override this.Child =
```
vasily-kirichenko: as above.

src/FSharpVSPowerTools.Logic/Linting/LintUtils.fs
```diff
@@ -0,0 +1,194 @@
...
+    let getInitialLoadedConfigs dte =
+        updateLoadedConfigs dte LoadedConfigs.Empty
+
+    let private directorySeparator = Path.DirectorySeparatorChar.ToString()
+
+    let rec private listStartsWith = function
+    | (_, []) -> true
+    | (x::list, y::startsWithList) when x = y ->
+        listStartsWith (list, startsWithList)
+    | _ -> false
```
vasily-kirichenko: Could you indent the body?

src/FSharpVSPowerTools.Logic/Linting/LintUtils.fs
```diff
@@ -0,0 +1,194 @@
...
+            { Update = IgnoreFilesUpdate.Overwrite
+              Files = []
+              Content = viewModel.IgnoreFiles.IgnoreFiles 
+                |> String.concat Environment.NewLine } |> Some
+          Analysers = 
+            viewModel.Rules
+                |> Seq.map ruleViewModelToAnalyser
+                |> Seq.toList
+                |> (fun x -> (ruleViewModelToHintAnalyser viewModel.Hints)::x)
+                |> Map.ofList }
```
vasily-kirichenko: Could you rearrange it like this?
```fsharp
viewModel.Rules
|> Seq.map ruleViewModelToAnalyser
|> Seq.toList
|> (fun x -> (ruleViewModelToHintAnalyser viewModel.Hints)::x)
|> Map.ofList }
```

dungpa:
Sorry that I haven't followed status of this PR close enough.

Is this PR ready for merging (aside from the current failing test)?

duckmatt:
Before it's merged I think it needs a message prompt dialog on saving changes to notify the user if the save was successful (could have failed due to no write access), this is something I'm starting to look into

dungpa:
Thanks for the update.

forki:
Are the settings stored per repo?

vasily-kirichenko:
I think they are stored per project in files and may be commited to VCS.

duckmatt:
Yep, it tries to save the settings to the solution directory by default (and these settings can be overridden with settings files in project directories for project specific settings)

duckmatt:
There's now a dialog that lets the user retry or cancel a save when their save fails. The failing test is now failing with:

    Attempt by method 'FSharpLint.Framework.HintParser+Identifiers+pidenttext@118-1.Invoke(Microsoft.FSharp.Collections.FSharpList`1<Char>)' 
    to access method 'Microsoft.FSharp.Compiler.Lexhelp+Keywords.get_keywordNames()' failed.

I'll try to release a new version of the FSharpLint api in line with the latest FSharp.Compiler.Service to fix this

dungpa:
The failing test has been fixed.

@vasily-kirichenko Please merge this PR if you think it's ready.

vasily-kirichenko:
I'm not sure what does it means:
image
The group is selected, but no one child is. It's confusing.
Either this one:
image
As the group is not checked, I'm not sure the selected child will work.
My propose: when user checks a group, all children should be checked automatically. And vise versa: if you uncheck a group, then all the children should be unchecked. It opens a new question: what state should a group be in if some children are checked, but others are not. It seems a common solution is to introduce the third group state for this:

vasily-kirichenko:
If I want to disable some rule, it's hard to find the setting related to it. For example, I've no idea how to switch it off:

image

vasily-kirichenko:
It would be more comfortable if the checkbox were checked/unchecked by clicking on "Enabled" label, too:

image

vasily-kirichenko:
1. Disable some rule that's enabled by default.
2. The config file has the settings off:
```xml
<?xml version="1.0" encoding="utf-8"?>
<FSharpLintSettings xmlns="https://github.com/fsprojects/FSharpLint/blob/master/ConfigurationSchema.xsd">
  <IgnoreFiles Update="Overwrite"><![CDATA[assemblyinfo.*]]></IgnoreFiles>
  <Analysers>
    <Binding>
      <Rules>
        <WildcardNamedWithAsPattern>
          <Enabled>True</Enabled>
        </WildcardNamedWithAsPattern>
      </Rules>
    </Binding>
    <Hints>
      <Rules />
      <Enabled>False</Enabled>
    </Hints>
  </Analysers>
</FSharpLintSettings>
```
3. Turn the rule on back.
4. The settings file contains the settings with default value:
```xml
<?xml version="1.0" encoding="utf-8"?>
<FSharpLintSettings xmlns="https://github.com/fsprojects/FSharpLint/blob/master/ConfigurationSchema.xsd">
  <IgnoreFiles Update="Overwrite"><![CDATA[assemblyinfo.*]]></IgnoreFiles>
  <Analysers>
    <Binding>
      <Rules>
        <WildcardNamedWithAsPattern>
          <Enabled>False</Enabled>
        </WildcardNamedWithAsPattern>
      </Rules>
    </Binding>
    <Hints>
      <Rules />
      <Enabled>False</Enabled>
    </Hints>
  </Analysers>
</FSharpLintSettings>
```
Expected behavior: default settings should not be stored in the settings file.

duckmatt:
I agree with all your points. 

I think the solution to finding a setting for a warning will be to introduce identifiers for the lint suggestions

The last point is a difficult one, the reason why the default settings are left in a file is in case someone wanted to enforce a default - this would be a rare case where someone could have a default setting changed in a setting file in an ancestor directory. I think as this case is somewhat rare it might be worth not adding default values, and if someone wants that behaviour they can edit the file themselves, what do you think?

vasily-kirichenko:
OK, I agree about default settings left in the files, let's leave it as it is.

About identifiers for suggestions, it'd work, but not that elegant. A good solution would be to add smart tags, which would include something like "switch this rule off for current project / whole solution", but it requires a lot of work. However, if we plan to add quick actions for suggestions, adding one more action into them would be quite easy.

dungpa:
I think adding labels for FSharpLint warnings is a good solution.

As a user, I would prefer:

```
FL0001: Identifiers Must Not Contain Underscores
```

than

```
IdentifiersMustNotContainUnderscores
```

in the setting dialog.

It's easy to understand and look up for further documentation.

We can always add smart tags support if we get to it later.

vasily-kirichenko:
@dungpa :+1:

duckmatt:
Great, I'll look at making those UI changes, and adding support to FSharpLint for the rule identifiers

vasily-kirichenko:
The checkboxes behave better now, but still inconsistently:

duckmatt:
Thanks for pointing that out - should be sorted now. I was looking at doing the tri-state checkboxes but thinking more about it I don't think it fits perfectly as it's possible for the analysers (the parent checkboxes) to only have a disabled/enabled state in the config file

vasily-kirichenko:
Great job, Matt! Thanks a lot both for FsLinter and for this integration.

dungpa:
@duckmatt Thank you for the hard work and patience on this PR.

Would you mind to add a short documentation page with screenshots for the feature? It's the last thing we hope to finish before releasing VFPT 2.1.0.

forki:
Great stuff. Kudos

vasily-kirichenko:
@dungpa I think we could include the outlining feature as well.

duckmatt:
Thanks guys, I'll look at getting a doc page together

PR #1100, opened 2015-08-31
Changing open document forces FCS to stop current checking

vasily-kirichenko:


dkinitiate:
A few notes:

spelling: absolete -> obsolete
Is it possible to substitute `"%HOMEDRIVE%%HOMEPATH%"` -> `"%USERPROFILE%"` ?
redundant binding `newState` in `__.FileChanged`
FileChanged, FileClosed are named like events, consider MarkFileAs*

Not a big fan of duplication of data : `LanguageService.files` and the open file `Map` in `OpenDocumentTracker`

vasily-kirichenko:
> Is it possible to substitute "%HOMEDRIVE%%HOMEPATH%" -> "%USERPROFILE%" ?

I don't know. It seems to be a fake path for cases where we cannot get full path for a file. 

> Not a big fan of duplication of data : LanguageService.files and the open file Map in OpenDocumentTracker

I agree, but I heavily use `ConcurrentDictionary` api here. One option is to make  `OpenDocumentTracker` use it, too, then move `FileState` there.

enricosada:
> Is it possible to substitute "%HOMEDRIVE%%HOMEPATH%" -> "%USERPROFILE%" ?
> I don't know. It seems to be a fake path for cases where we cannot get full path for a file.

not always `%HOMEDRIVE%%HOMEPATH%` = `%USERPROFILE%`, can be changed with group policy ( at work i had mine mapped to a network drive )

- `%HOMEDRIVE%%HOMEPATH%`: usually is the the user personal files, my documents, music etc
- `%USERPROFILE%`: user config files (roaming or local) etc

i usually use this [msdn article](http://blogs.msdn.com/b/patricka/archive/2010/03/18/where-should-i-store-my-data-and-configuration-files-if-i-target-multiple-os-versions.aspx) to choose where store config files

PR #1101, opened 2015-08-31
Tightened up SourceCodeClassifier

cloudRoutine:
`( Seq.map |> Seq.concat ) = Seq.Collect`
added `List.skipWhile` to drop unnecessary conversions
`Seq.Filter` preceding `Seq.Fold` can be fused into the fold


dungpa:
Look like the right things to do. Thanks for looking to it.

Do you recognize any performance improvement (e.g. reducing amount of GC) after the changes?

vasily-kirichenko:
Looks cool, thanks!

Maybe we should consider switching to Nessos Streams everywhere in the codebase?

cloudRoutine:
I tried out streams and they're not a good fit for how the code is currently structured. Using them would create a lot more unnecessary allocations and the many short lived streams that can't be chained/composed easily are a huge performance hamper. Not to mention streams don't support all the transformation functions used and a shift from Stream toSeq or toArray seems a lot more expensive than Seq to List and vice versa.

@dungpa Honestly the performance gains were minuscule at best in most cases. There's a tiny shift to more gen 0 collections than gen 1, but once the file size is up to 10k+ it helps a bit.

vasily-kirichenko:
I just recall that Nessos Streams perform well (very, very well) with x64 only, so no luck for us.

PR #1103, opened 2015-09-01
Deforesting

cloudRoutine:
map |> concat -> collect
filter |> map    -> fold

### DIFF
--- a/FSharpVSPowerTools.sln
+++ b/FSharpVSPowerTools.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.23107.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{A6A6AF7D-D6E3-442D-9B1E-58CC91879BE1}"
 EndProject
@@ -78,6 +78,9 @@ EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "FSharpVSPowerTools.Logic.VS2015", "src\FSharpVSPowerTools.Logic.VS2015\FSharpVSPowerTools.Logic.VS2015.fsproj", "{BB0C6F01-5E57-4575-A498-5DE850D9FA6C}"
 EndProject
 Global
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Debug|Mixed Platforms = Debug|Mixed Platforms
--- a/FSharpVSPowerTools.sln
+++ b/FSharpVSPowerTools.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+# Visual Studio 2013
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{A6A6AF7D-D6E3-442D-9B1E-58CC91879BE1}"
 EndProject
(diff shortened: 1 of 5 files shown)


src/FSharpVSPowerTools.Core/OpenDeclarationsGetter.fs
```diff
@@ -46,23 +46,22 @@ module OpenDeclarationGetter =
 
     let private getAutoOpenModules entities =
         entities 
-        |> List.filter (fun e -> 
+        |> List.fold( fun acc e ->
```
dungpa: I think in order to preserve ordering, `foldBack` should be used instead of `fold`.

One could use `choose` instead of filter + map, but `choose` need quite many allocations for temporary option values.

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -9,7 +9,7 @@ type internal ShortIdent = string
 type internal Idents = ShortIdent[]
 
 let internal longIdentToArray (longIdent: LongIdent): Idents =
-    longIdent |> List.map string |> List.toArray
+    Array.init longIdent.Length (fun i -> string longIdent.[i])
```
dungpa: I don't think this is desirable. `List.length` is `O(n)`. Combining with index access `.[i]`, the whole `Array.init` call would be of `O(n^2)` where n is the length of the list.

If we want to avoid allocation, changing to

```fsharp
Seq.map string |> Seq.toArray
```

would be enough

vasily-kirichenko: I agree with @dungpa, except that `longIdent.Length` is called only once. The real problem here is `longIdent.[i]` which is O(n).

FSharpVSPowerTools.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.23107.0
```
vasily-kirichenko: If you like to use VS 2015, open `FSharpVSPowerTools.v14.sln`.

FSharpVSPowerTools.sln
```diff
@@ -78,6 +78,9 @@ EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "FSharpVSPowerTools.Logic.VS2015", "src\FSharpVSPowerTools.Logic.VS2015\FSharpVSPowerTools.Logic.VS2015.fsproj", "{BB0C6F01-5E57-4575-A498-5DE850D9FA6C}"
 EndProject
 Global
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
```
vasily-kirichenko: Please remove these lines, they cause the performance toolbar appears every time one opens the solution.

src/FSharpVSPowerTools.Core/OpenDeclarationsGetter.fs
```diff
@@ -460,7 +457,6 @@ module OpenDeclarationGetter =
                     |> Option.bind (function
                          | [||] -> None
                          | prefix -> Some (sUseRange, prefix)))) 
-            |> Array.concat
```
cloudRoutine: I was thinking this could be changed to something like

        let private getFullPrefix (longIdents: IDictionary<_,_>) fullName (endPos: Range.pos): Idents option =
            match longIdents.TryGetValue endPos with
            | true, longIdent ->                   
                let rec loop matchFound (longIdents:_[]) (symbolIdents:_[]) (lcnt:int) (scnt:int) =
                    match lcnt, scnt with
                    | lcnt, _    when lcnt = longIdents.Length    -> symbolIdents.[scnt..]
                    | _   , scnt when scnt = symbolIdents.Length  -> [||]
                    | lcnt, scnt ->
                        if longIdents.[lcnt] <> symbolIdents.[scnt] then
                            if matchFound 
                            then symbolIdents.[scnt..] 
                            else loop matchFound longIdents symbolIdents (lcnt+1) (scnt+1)
                        else loop true longIdents symbolIdents (lcnt+1) (scnt+1)
                let prefix = 
                    loop false (longIdent |> Array.rev) (fullName |> Array.rev) 0 0
                    |> Array.rev

but it's probably a bit excessive to get 
```
(longIdent |> Array.rev |> List.ofArray) (fullName |> Array.rev |> List.ofArray)
|> List.rev
|> List.toArray
``` 
to become 
```
loop false (longIdent |> Array.rev) (fullName |> Array.rev) 0 0
|> Array.rev
```

dungpa: I think it is probably fine as it is now.

src/FSharpVSPowerTools.Core/OpenDeclarationsGetter.fs
dungpa: @cloudRoutine Could you change this to `foldBack` as well for consistency?

dungpa:
@cloudRoutine Thanks a lot.

PR #1108, opened 2015-09-06
Setting cursor positions before creating taggers

### LINKED ISSUE #1107: VS unit tests fail on master
After #1106, getting cursor positions moves to ThreadPool. 

Setting cursor position might occur after the first TagChanged event, therefore, test assertions fail.

These unit tests need to be updated.

dungpa:
Close #1107.


vasily-kirichenko:
Great, thanks!

PR #1109, opened 2015-09-06
Read all F# projects recursively for generating references

### LINKED ISSUE #1090: load-references does not auto-regenerate in VS2015
Manually calling "Generate references for F# Interactive" works fine - it just doesn't happen on building solutions any more.

dungpa:
Close #1090.


vasily-kirichenko:
Great, works OK on my solution.

PR #1117, opened 2015-09-17
"Go To Definition" to handle #load directives

smoothdeveloper:
I tested those changes on VS2015 and it seems to work, I had issues with VS2013 on one machine but they seem to be related to something else going on in my VS experimental instance (see commit message https://github.com/smoothdeveloper/VisualFSharpPowerTools/commit/094de24ac5419ae8262dee8ac81741c5ddbdf472)

Would appreciate if someone can try on other machines with VS2013, just load a .fsx which has #load directives and try "Go To Definition" on it.

Please let me know if you see any change in the code to be made.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -52,7 +52,7 @@ let msBuildVersionToTry = ["12.0";"14.0"]
 
 let maybeVersion =
     msBuildVersionToTry
-    |> Seq.map (fun v -> ProgramFilesX86 @@ (sprintf @"\MSbuild\%s\Bin\MSbuild.exe"))
+    |> Seq.map (fun v -> ProgramFilesX86 @@ (sprintf @"\MSbuild\%s\Bin\MSbuild.exe" v))
     |> Seq.tryFind File.Exists
     
 match maybeVersion with
--- a/build.fsx
+++ b/build.fsx
@@ -47,9 +47,6 @@ let gitHome = "https://github.com/" + gitOwner
 let gitName = "VisualFSharpPowerTools"
 let cloneUrl = "git@github.com:fsprojects/VisualFSharpPowerTools.git"
 
-// Ensure to use MSBuild 12 for the main solution if it exists, otherwise use version 14
-let msBuildVersionToTry = ["12.0";"14.0"]
-
 // Default to MSBuild 12 if present
 let msBuildPath = (ProgramFilesX86 @@ @"\MSBuild\12.0\Bin\MSBuild.exe")
 if File.Exists msBuildPath then
--- a/build.fsx
+++ b/build.fsx
@@ -50,14 +50,10 @@ let cloneUrl = "git@github.com:fsprojects/VisualFSharpPowerTools.git"
 // Ensure to use MSBuild 12 for the main solution if it exists, otherwise use version 14
 let msBuildVersionToTry = ["12.0";"14.0"]
 
-let maybeVersion =
-    msBuildVersionToTry
-    |> Seq.map (fun v -> ProgramFilesX86 @@ (sprintf @"\MSbuild\%s\Bin\MSbuild.exe" v))
-    |> Seq.tryFind File.Exists
-    
-match maybeVersion with
-| None -> failwithf "couldn't find MSBuild"
-| Some msBuildPath -> setEnvironVar "MSBuild" msBuildPath
+// Default to MSBuild 12 if present
+let msBuildPath = (ProgramFilesX86 @@ @"\MSBuild\12.0\Bin\MSBuild.exe")
+if File.Exists msBuildPath then
+    setEnvironVar "MSBuild" msBuildPath
 
 // Read additional information from the release notes document
 Environment.CurrentDirectory <- __SOURCE_DIRECTORY__
--- a/build.fsx
+++ b/build.fsx
@@ -47,8 +47,17 @@ let gitHome = "https://github.com/" + gitOwner
 let gitName = "VisualFSharpPowerTools"
 let cloneUrl = "git@github.com:fsprojects/VisualFSharpPowerTools.git"
 
-// Ensure to use MSBuild 12 for the main solution
-setEnvironVar "MSBuild" (ProgramFilesX86 @@ @"\MSBuild\12.0\Bin\MSBuild.exe")
+// Ensure to use MSBuild 12 for the main solution if it exists, otherwise use version 14
+let msBuildVersionToTry = ["12.0";"14.0"]
+
+let maybeVersion =
+    msBuildVersionToTry
+    |> Seq.map (fun v -> ProgramFilesX86 @@ (sprintf @"\MSbuild\%s\Bin\MSbuild.exe"))
+    |> Seq.tryFind File.Exists
+    
+match maybeVersion with
+| None -> failwithf "couldn't find MSBuild"
+| Some msBuildPath -> setEnvironVar "MSBuild" msBuildPath
 
 // Read additional information from the release notes document
 Environment.CurrentDirectory <- __SOURCE_DIRECTORY__
--- a/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
@@ -749,16 +749,14 @@ module HashDirectiveInfo =
     /// returns the Some (complete file name of a resolved #load directive at position) or None
     let getHashLoadDirectiveResolvedPathAtPosition (pos: pos) (ast: ParsedInput) : string option =
         let l = pos.Line
-        let directive =
-            getIncludeAndLoadDirectives ast
-            |> Array.tryFind (function
-                | Load(ExistingFile _, range)
-                  // check the line is within the range
-                  // (doesn't work when there are multiple files given to a single #load directive)
-                  when range.StartLine <= l && range.EndLine >= l
-                    -> true
-                | _ -> false
-            )
-        match directive with
-        | Some ( Load( ExistingFile f, _) ) -> Some f
-        | _ -> None
+
+        getIncludeAndLoadDirectives ast
+        |> Array.tryPick (
+            function
+            | Load(ExistingFile f, range) 
+                // check the line is within the range
+                // (doesn't work when there are multiple files given to a single #load directive)
+                when range.StartLine <= l && range.EndLine >= l
+                    -> Some f
+            | _     -> None
+        )
--- a/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
@@ -748,8 +748,6 @@ module HashDirectiveInfo =
 
     /// returns the Some (complete file name of a resolved #load directive at position) or None
     let getHashLoadDirectiveResolvedPathAtPosition (pos: pos) (ast: ParsedInput) : string option =
-        let l = pos.Line
-
         getIncludeAndLoadDirectives ast
         |> Array.tryPick (
             function
--- a/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
@@ -756,7 +756,7 @@ module HashDirectiveInfo =
             | Load(ExistingFile f, range) 
                 // check the line is within the range
                 // (doesn't work when there are multiple files given to a single #load directive)
-                when range.StartLine <= l && range.EndLine >= l
+                when rangeContainsPos range pos
                     -> Some f
             | _     -> None
         )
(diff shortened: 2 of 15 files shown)


tests/FSharpVSPowerTools.Core.Tests/GoToDefinitionTests.LoadDirective.fs
```diff
@@ -0,0 +1,80 @@
...
+let checker = FSharpChecker.Create()
+
+let canonicalizeFilename filename = Path.GetFullPathSafe filename //(new FileInfo(filename)).FullName
+
+let getAst filename = 
+    let contents = File.ReadAllText(filename)
+
+      // Get compiler options for the 'project' implied by a single script file
+    let projOptions = 
+        checker.GetProjectOptionsFromScript(filename, contents)
```
dungpa: I think we should use `LanguageService` instead of using `FSharpChecker` directly. Most of out core tests rely on `LanguageService`.

In `LanguageService.GetScriptCheckerOptions`, you could use `FSharpCompilerVersion.FSharp_3_1` to ensure compatibility with F# 3.1.

tests/FSharpVSPowerTools.Tests/FSharpVSPowerTools.Tests.fsproj
```diff
@@ -97,6 +97,8 @@
     <Compile Include="ImplementInterfaceSmartTaggerTests.fs" />
     <Compile Include="GoToDefinitionFilterCommandTests.fs" />
     <Compile Include="LintTaggerTests.fs" />
+    <None Include="Scripts\load-references.fsx" />
```
dungpa: Do we need include fsx files for this

tests/FSharpVSPowerTools.Core.Tests/GoToDefinitionTests.LoadDirective.fs
```diff
@@ -0,0 +1,80 @@
...
+   
+   let results =
+       directives
+       |> Seq.map (function
+          | Load(ExistingFile(filename), _) -> Some ((new FileInfo(filename)).FullName)
+          | _ -> None
+       )
+       |> Seq.toList
+   
+   Assert.AreEqual(expectedMatches, results)
```
dungpa: Please use type-safe `assertEqual` instead https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/tests/FSharpVSPowerTools.Core.Tests/TestHelpers.fs#L35.

tests/FSharpVSPowerTools.Core.Tests/GoToDefinitionTests.LoadDirective.fs
```diff
@@ -0,0 +1,80 @@
...
+       |> Seq.map fst
+       |> Seq.map (fun pos -> 
+          let result = getHashLoadDirectiveResolvedPathAtPosition pos ast
+          match result with
+          | None      -> pos, None
+          | Some path -> pos, Some (canonicalizeFilename path)
+       )
+       |> Seq.toList
+   
+   Assert.AreEqual(expectations, results)
```
dungpa: Same comment as above.

tests/FSharpVSPowerTools.Tests/FSharpVSPowerTools.Tests.fsproj
smoothdeveloper: Sorry removing this one.

tests/FSharpVSPowerTools.Core.Tests/GoToDefinitionTests.LoadDirective.fs
smoothdeveloper: Will try it out and change to use `LanguageService` makes more sense.

build.fsx
```diff
@@ -47,8 +47,17 @@ let gitHome = "https://github.com/" + gitOwner
 let gitName = "VisualFSharpPowerTools"
 let cloneUrl = "git@github.com:fsprojects/VisualFSharpPowerTools.git"
 
-// Ensure to use MSBuild 12 for the main solution
-setEnvironVar "MSBuild" (ProgramFilesX86 @@ @"\MSBuild\12.0\Bin\MSBuild.exe")
+// Ensure to use MSBuild 12 for the main solution if it exists, otherwise use version 14
+let msBuildVersionToTry = ["12.0";"14.0"]
```
dungpa: Thanks for fixing this. I think the logic could be simplified as follows:

Check if MsBuild 12 exists and set priority to MsBuild 12. Otherwise, do nothing.

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -653,3 +653,112 @@ let getModuleOrNamespacePath (pos: pos) (ast: ParsedInput) =
...
+    /// returns the Some (complete file name of a resolved #load directive at position) or None
+    let getHashLoadDirectiveResolvedPathAtPosition (pos: pos) (ast: ParsedInput) : string option =
+        let l = pos.Line
+        let directive =
+            getIncludeAndLoadDirectives ast
+            |> Array.tryFind (function
+                | Load(ExistingFile _, range)
+                  // check the line is within the range
+                  // (doesn't work when there are multiple files given to a single #load directive)
+                  when range.StartLine <= l && range.EndLine >= l
```
dungpa: I think we should use posInRange here (see https://github.com/fsprojects/VisualFSharpPowerTools/blob/c474ac16df14d04b820f8296d83a47439867ca4a/src/FSharpVSPowerTools.Core/InterfaceStubGenerator.fs#L575)

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -653,3 +653,112 @@ let getModuleOrNamespacePath (pos: pos) (ast: ParsedInput) =
...
+                    -> true
+                | _ -> false
+            )
+        match directive with
```
dungpa: If you use `Array.tryPick`, you don't have to pattern match again here.

src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
```diff
@@ -46,13 +46,24 @@ type GoToDefinitionFilter(textDocument: ITextDocument,
     let urlChanged = if fireNavigationEvent then Some (Event<UrlChangeEventArgs>()) else None
     let mutable currentUrl = None
 
+    let getCurrentFilePathProjectAndDoc () =
+        maybe {
+            let filepath = textDocument.FilePath
+            let dte = serviceProvider.GetService<EnvDTE.DTE, SDTE>()
+            let! doc = dte.GetCurrentDocument(filepath)
+            let! project = maybe {
```
dungpa: This `maybe` computation expression block is entirely redundant. We can move it to the top level.

src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
```diff
@@ -421,10 +432,21 @@ type GoToDefinitionFilter(textDocument: ITextDocument,
                 match symbolResult with
                 | Some (_, _, _, _, FSharpFindDeclResult.DeclFound _) 
                 | None ->
-                    // Run the operation on UI thread since continueCommandChain may access UI components.
-                    do! Async.SwitchToContext uiContext
-                    // Declaration location might exist so let Visual F# Tools handle it. 
-                    return continueCommandChain()
+                    // no FSharpSymbol found, here we look at #load directive
```
dungpa: The logic here is a bit convoluted. If we modify getDocumentState to return:
```fsharp
type Result =
| DeclFound of ...
| DeclNotFound of ...
| NoResult of filePath:string * project:IProjectProvider
```

then we can remove an extra call to `getCurrentFilePathProjectAndDoc`.

Moreover, `Async.RunSynchronously` could easily be removed (We generally avoid it in the code base).

src/FSharpVSPowerTools.Logic/VSUtils.fs
```diff
@@ -87,6 +87,20 @@ type ITextBuffer with
...
 
+type ITextView with
+    member x.GetCaretPosition () =
+        maybe {
+          let! point = x.TextBuffer.GetSnapshotPoint x.Caret.Position
+          let line = point.Snapshot.GetLineNumberFromPosition point.Position
+          let col = point.Position - point.GetContainingLine().Start.Position
+          return (line, col)
+        }
+    member x.posAtCaretPosition () =
```
dungpa: This member should be PascalCase.

dungpa:
The PR looks very good. I tested on a machine with only VS 2013. It works fine; there is an issue with imprecise range.

src/FSharpVSPowerTools.Logic/VSUtils.fs
smoothdeveloper: I guess I put it to avoid matching on GetSnapshotPoint.

Would you recommend something like this:
```fsharp
member x.GetCaretPosition () =
    match x.TextBuffer.GetSnapshotPoint x.Caret.Position with
    | None -> None
    | Some point ->
        let line = point.Snapshot.GetLineNumberFromPosition point.Position
        let col = point.Position - point.GetContainingLine().Start.Position
        Some (line, col)
```
or I'm missing something?

dungpa: I meant to follow naming convention: `posAtCaretPosition` --> `PosAtCaretPosition`. Is this method used at all?

build.fsx
smoothdeveloper: Good point, I was maybe looking to far ahead (up to msbuid 22 :)) will make it a simpler condition as you suggest.

src/FSharpVSPowerTools.Logic/VSUtils.fs
smoothdeveloper: Method is used in VSLanguage.GetLoadDirectiveFileNameAtCursor.

src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
smoothdeveloper: I fixed all other remarks.

I'll give a look at this tomorrow. I didn't change ```getDocumentState``` (beside extracting some code to get filepath, project and document which is also used with the changes in ```gotoDefinition```) and would certainly see changing it's return type be relevant but felt like too much overhead for my change when I was fixing ```goToDefinition```.

I'll spend time tomorrow and will implement the refactoring you propose, thanks for giving me a clue about expected return type!

smoothdeveloper: Regarding `Async.RunSynchronously` I'm trying to remove it:

````fsharp
                    // no FSharpSymbol found, here we look at #load directive
                    let directive = maybe {
                        let! filepath, project, _ = getCurrentFilePathProjectAndDoc()
                        return vsLanguageService.GetLoadDirectiveFileNameAtCursor(filepath, view, project)
                    }
                    let maybeFileToOpen =
                        match directive with
                        | Some maybeResultAsync ->
                            let! r = maybeResultAsync // This construct may only be used within computation expressions
                            r
                        | None -> None

                    match maybeFileToOpen with
                    | Some fileToOpen ->
                        // directive found, navigate to the file at first line
                        serviceProvider.NavigateTo(fileToOpen, 0, 0, 0, 0)
                    | None ->
                        // Run the operation on UI thread since continueCommandChain may access UI components.
                        do! Async.SwitchToContext uiContext
                        // Declaration location might exist so let Visual F# Tools handle it. 
                        return continueCommandChain()
```

although the whole thing is within `async { ... }` and not in another computation expression.

Is this me doing something incorrect or a compiler bug?

dungpa: Different computation expressions don't compose well.

You should use `asyncMaybe` monad. 
I think this should works:

```fsharp
let directive = 
    asyncMaybe {
         let! filepath, project, _ = getCurrentFilePathProjectAndDoc()
         return! vsLanguageService.GetLoadDirectiveFileNameAtCursor(filepath, view, project) |> liftAsync
      }
match directive with
| Some fileToOpen -> ...
| None -> ...
```

See another example at https://github.com/fsprojects/VisualFSharpPowerTools/blob/199bf001fab7d1bea0119e2661f34326d68db72a/src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs#L90-L132

smoothdeveloper: Got it thanks!

I'm trying to see if I should remove `Async.RunSynchronously` I've put in `GoToDefinitionTests.LoadDirective.fs` `getAst`, is it something we are concerned with?

dungpa: Oh, we use `Async.RunSynchronously` in the tests all the time. Please leave it as is.

dungpa:
Thanks for the great work.

smoothdeveloper:
@vasily-kirichenko / @dungpa thanks for all the assistance and thorough review.

Here is a small preview of the feature:

gotodefinition on load directive

It should be noted that there is limitation in handling `#load` directive with multiple files, in which case it will jump to the first one.

I'll study how to address that concern but please consider merging this PR if the changes look right.

PR #1123, opened 2015-09-27
Split lines refactoring

smoothdeveloper:
This replaces direct usages of String.Split whenever the intent was to get an array of lines.

Introduced two functions in Utils.fs:

* `String.getLines`
* `String.getNonEmptyLines`

the only non-obvious thing which happens is that in first function, an empty line has to be yielded at the end if the input string ends with a trailing linebreak, this is due to behaviour of `StringReader`.

I stepped through all the changes in debug, but couldn't step into those two places:
* `DepthParser.GetRange` I believe this method is not used anywhere, we might want to eventually delete it
* `XmlDocParser.GetXmlDocables` please let me know if you have a way to trigger this code, I tried adding comments after triple slashes and compiling a project, then using the documented member but couldn't step there

### DIFF
--- a/src/FSharpVSPowerTools.Core/SignatureGenerator.fs
+++ b/src/FSharpVSPowerTools.Core/SignatureGenerator.fs
@@ -843,7 +843,7 @@ and internal writeDocs writer docs getXmlDocSig xmlDocBySig =
         else docs :> seq<_>
 
     xmlDocs
-    |> Seq.collect (fun line -> line.Replace("\r\n", "\n").Split('\r', '\n'))
+    |> Seq.collect (String.getLines)
     |> Seq.iter (fun line -> writer.WriteLine("/// {0}", line.TrimStart()))
 
 and internal writeActivePattern ctx (case: FSharpActivePatternCase) =
--- a/src/FSharpVSPowerTools.Core/SignatureGenerator.fs
+++ b/src/FSharpVSPowerTools.Core/SignatureGenerator.fs
@@ -843,7 +843,7 @@ and internal writeDocs writer docs getXmlDocSig xmlDocBySig =
         else docs :> seq<_>
 
     xmlDocs
-    |> Seq.collect (String.getLines)
+    |> Seq.collect String.getLines
     |> Seq.iter (fun line -> writer.WriteLine("/// {0}", line.TrimStart()))
 
 and internal writeActivePattern ctx (case: FSharpActivePatternCase) =
--- a/src/FSharpVSPowerTools.Core/Utils.fs
+++ b/src/FSharpVSPowerTools.Core/Utils.fs
@@ -420,7 +420,28 @@ module String =
         elif value.Contains(pattern) then
             Some value
         else None
-     
+    
+    open System.IO
+
+    let getLines (str: string) =
+        use reader = new StringReader(str)
+        [|
+        let line = ref (reader.ReadLine())
+        while (!line |> function | null -> false | _ -> true) do
+            yield !line
+            line := reader.ReadLine()
+        |]
+
+    let getNonEmptyLines (str: string) =
+        use reader = new StringReader(str)
+        [|
+        let line = ref (reader.ReadLine())
+        while ( !line |> function | null -> false | _ -> true ) do
+            if (!line).Length > 0 then
+                yield !line
+            line := reader.ReadLine()
+        |]
+
 [<AutoOpen; CompilationRepresentation (CompilationRepresentationFlags.ModuleSuffix)>]
 module Pervasive =
     open System.Diagnostics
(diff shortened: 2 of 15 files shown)


src/FSharpVSPowerTools.Core/SignatureGenerator.fs
```diff
@@ -843,7 +843,7 @@ and internal writeDocs writer docs getXmlDocSig xmlDocBySig =
         else docs :> seq<_>
 
     xmlDocs
-    |> Seq.collect (fun line -> line.Replace("\r\n", "\n").Split('\r', '\n'))
+    |> Seq.collect (String.getLines)
```
dungpa: The parentheses are redundant.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -420,7 +420,32 @@ module String =
...
         else None
-     
+    
+    open System.IO
+
+    let getLines (str: string) =
+        use reader = new StringReader(str)
+        [|
+        let line = ref (reader.ReadLine())
+        while (!line |> function | null -> false | _ -> true) do
```
dungpa: What about `!line <> null`?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -420,7 +420,32 @@ module String =
...
+            // last trailing space not returned
+            // http://stackoverflow.com/questions/19365404/stringreader-omits-trailing-linebreak
+            yield String.Empty
+        |]
+
+    let getNonEmptyLines (str: string) =
+        use reader = new StringReader(str)
+        [|
+        let line = ref (reader.ReadLine())
+        while ( !line |> function | null -> false | _ -> true ) do
```
dungpa: Same as above.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -420,7 +420,32 @@ module String =
...
+        while (!line |> function | null -> false | _ -> true) do
+            yield !line
+            line := reader.ReadLine()
+        if str.EndsWith("\n") then
+            // last trailing space not returned
+            // http://stackoverflow.com/questions/19365404/stringreader-omits-trailing-linebreak
+            yield String.Empty
+        |]
+
+    let getNonEmptyLines (str: string) =
```
dungpa: What do you think of introducing a private function that extract commonalities of these two functions using a boolean flag? It will be better in keeping two similar implementations in sync.

src/FSharpVSPowerTools.Core/Utils.fs
vasily-kirichenko: Comparing to `null` is very inefficient. If I remember properly, @latkin made benchmarks. PM is ok.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -420,7 +420,32 @@ module String =
...
             Some value
         else None
-     
+    
+    open System.IO
+
+    let getLines (str: string) =
+        use reader = new StringReader(str)
+        [|
+        let line = ref (reader.ReadLine())
```
vasily-kirichenko: Maybe a mutable var would be better?

dungpa:
The PR looks good. It would be great to update these tests using new getLines* functions https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/tests/FSharpVSPowerTools.Core.Tests/TaskListCommentExtractorTests.fs.

src/FSharpVSPowerTools.Core/Utils.fs
smoothdeveloper: The code was with mutable but had to change it to ref for compiling with VS2013.

The value is mutated in a closure.

src/FSharpVSPowerTools.Core/Utils.fs
dungpa: Bummer. http://latkin.org/blog/2015/05/18/null-checking-considerations-in-f-its-harder-than-you-think/

I guess I complained because using `function` is a bit weird here. Perhaps `match !line with` is more natural?

smoothdeveloper: I agree it's confusing, I'm adding isNull and isNotNull inline functions in Pervasive (Utils.fs) and moving the String module bellow the Pervasive one.

src/FSharpVSPowerTools.Core/Utils.fs
smoothdeveloper: I considered having one function based on other (`Array.filter`) but it seemed 20% slower.

Also I feel those low level short functions aren't going to cause maintenance trouble so I came up with most efficient and obvious implementation to me.

Let me know if you prefer a single base function which might perform slightly worse but not too drastic in which case I can change it.

vasily-kirichenko: It sounds weird. It must be tens or hundreds of millions calls per second to make function call a performance bottleneck. 

Could you profile it?

If this is the case, then try to make the function `inline`.

dungpa: I think the explanation is OK; the individual functions are small enough anyway.

dungpa:
Thanks for the thorough updates.

Issue #1126, opened 2015-09-30
Function type alias is not colored


OkayX6:
At definition: 
![image](https://cloud.githubusercontent.com/assets/1169339/10197287/17016dc6-6797-11e5-93ca-51ee4344f6c3.png)

At usage:
![image](https://cloud.githubusercontent.com/assets/1169339/10197304/2aba3dfc-6797-11e5-9c51-9a0beb1df413.png)

vasily-kirichenko:
Is it colored as function?

OkayX6:
No :)

vasily-kirichenko:
Is `unit -> unit` a type?

vasily-kirichenko:
Fixed in https://github.com/fsprojects/VisualFSharpPowerTools/pull/1129

PR #1129, opened 2015-09-30
color function type aliases as reference types

vasily-kirichenko:
Fix for https://github.com/fsprojects/VisualFSharpPowerTools/issues/1126

### DIFF
--- a/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
+++ b/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
@@ -179,11 +179,14 @@ module private OperatorCategorizer =
 
 module SourceCodeClassifier =
     let getIdentifierCategory = function
-        | Entity (_, ValueType, _) -> Category.ValueType
-        | Entity Class -> Category.ReferenceType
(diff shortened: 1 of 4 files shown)


dungpa:
@OkayX6 Do you have time to review this :-)?

OkayX6:
Yeah, so the code looks good and I will try the extension :)

vasily-kirichenko:
Maybe it's better to colorize it as Function? Does it make sense?

dungpa:
I think coloring it as a reference type is right.
```fsharp
type Action = unit -> unit
let f : action = failwith ""
```

`Action` is a type, either a reference type or a value type.
`f` is a value of type `Action`; it may be colored as Function.

vasily-kirichenko:
Isn't `f` colored as Function already?

dungpa:
Just checked. Actually no, we haven't checked for function types that are indirect through type alias.

vasily-kirichenko:
OK, I'll fix it now.

OkayX6:
Cool. Other than that the update works well for me.

vasily-kirichenko:
Done. Ready for merging.

PR #1134, opened 2015-10-02
update FCS to 1.4.0.6

vasily-kirichenko:


dungpa:
I think a VS unit test is failing deterministically. I have to check.

dungpa:
Oh, I think the latest FAKE version isn't source-indexed /cc @forki.

We use latest FAKE as source indexing unit tests (see https://github.com/fsprojects/VisualFSharpPowerTools/blob/c92f530f0386bca4990265a0b7f8f64a65821edb/tests/data/NavigateToSource/paket.references#L2), so the test fails at the moment.

Could you pin FAKE version or better add an old FakeLib.dll to https://github.com/fsprojects/VisualFSharpPowerTools/tree/c92f530f0386bca4990265a0b7f8f64a65821edb/tests/data/NavigateToSource to use as test cases?

forki:
It's not source indexed? Really?

dungpa:
Not entirely sure. Maybe the paths have been changed.

It's a bad idea to use bleeding-edge packages as unit tests anyway.

forki:
Mhm. I agree: you unittest against specific version. Paket's new group
feature allows you to restore a legacy version of paket.

That said: would love to hear if sourcelinks are broken. /cc @Camr0n

PR #1136, opened 2015-10-03
fixed: quotations were not colorized in static members

vasily-kirichenko:
Fix for 
image
After
image


vasily-kirichenko:
It's very straightforward, so I'm merging it.

PR #1139, opened 2015-10-03
Colorize active patterns application as functions (WIP)

vasily-kirichenko:
The idea is to form `DraftToken` of kind `Ident` for each token sequence of the following shape:

`(|.....any other tokens here....|)`

Problems:

* inner tokens, like `Foo` and `Bar` in `(|Foo|Bar|_|)` seem to disappear
* definition and application must processed separately, which requires typed ast

### DIFF
--- a/src/FSharpVSPowerTools.Core/Lexer.fs
+++ b/src/FSharpVSPowerTools.Core/Lexer.fs
@@ -26,10 +26,8 @@ type SymbolLookupKind =
 
 type internal DraftToken =
     { Kind: SymbolKind
-      Token: FSharpTokenInfo 
+      LeftColumn: int
       RightColumn: int }
-    static member Create kind token = 
-        { Kind = kind; Token = token; RightColumn = token.LeftColumn + token.FullMatchedLength - 1 }
 
 module Lexer =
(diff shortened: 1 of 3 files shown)


dungpa:
I think it's a good idea. But man, this is getting complex.

vasily-kirichenko:
Yep :( Another problem is that the artificial token messes the rest of features like this:


 
I'm going to give up on this.

dungpa:
Your call :-). I spent the whole last night on this, but didn't find a reasonable solution.

vasily-kirichenko:
Amen :)

dungpa:
What about adding a special case for active pattern usage? If it is active pattern usage, we return the ranges directly; otherwise, call `lexer.GetSymbolFromTokensAtLocation` as usual. It would be a low-risk change.

vasily-kirichenko:
Then Highlight usages will stop working for individual AP's cases.

vasily-kirichenko:
So this whole thing should work for coloring only and not for the rest of the features.

vasily-kirichenko:
It's a mess, really.

PR #1140, opened 2015-10-03
Highlight active pattern usage as functions

### LINKED ISSUE #1137: Color active patterns being used as functions
Example:
```fsharp
    let (|AAA|_|) _ = Some ()
    let _ = (|AAA|_|) ()
```

Currently, usage of active patterns isn't being colored at all.

The reason is that we have a sanity check to rule out faulty ranges at https://github.com/fsprojects/VisualFSharpPowerTools/blob/a2abbd756b4226b71fb03946cdc9def8b3278550/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs#L234-L255. Active patterns' ranges consist of ranges for `|AAA|_|`, so right columns are 3 character far off compared to those of normal identifiers. It's tricky to fix since the lexer often looks up by right columns.

We even have an ignored failing test for this https://github.com/fsprojects/VisualFSharpPowerTools/blob/a2abbd756b4226b71fb03946cdc9def8b3278550/tests/FSharpVSPowerTools.Core.Tests/SourceCodeClassifierTests.fs#L866-L873

Extracted from #1132.

dungpa:
Another attempt to fix #1137.

The changes are only limited to coloring; they don't affect any other feature.

### DIFF
--- a/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
+++ b/src/FSharpVSPowerTools.Core/SourceCodeClassifier.fs
@@ -235,10 +235,12 @@ module SourceCodeClassifier =
                     let r = fsSymbolUse.RangeAlternate
                     match fsSymbolUse.Symbol with
                     | MemberFunctionOrValue func when func.IsActivePattern && not fsSymbolUse.IsFromDefinition ->
+                        // Usage of Active Patterns '(|A|_|)' has the range of '|A|_|',
+                        // so we expand the ranges in order to include parentheses into the results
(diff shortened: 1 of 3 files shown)


vasily-kirichenko:
What a nice solution! I was stuck with the Lexer and did not see the bigger picture :(

One more detail, I think we should colorize the parenthesis, too:

image

It looks inconsistent with double quoted APs and functions.

dungpa:
Good idea. Parentheses are now included into active patterns' spans.

vasily-kirichenko:
Great, works perfectly.

PR #1141, opened 2015-10-04
Revert "move execution of all `updateAtCaretPosition` into ThreadPool"

vasily-kirichenko:
This reverts commit 80a2b10ea26fae685b1bc633f16daf1ed00fd518.

It 1. may cause VS UI thread dead locked 2. bring little value


dungpa:
:+1:

PR #1149, opened 2015-10-11
WIP: Put dependencies in groups

forki:


forki:
One test gets a time out. Any ideas?

vasily-kirichenko:
That one is timed out in 50% of times. I re-run the build.

vasily-kirichenko:
Ah, that's another one. We should pin FAKE to the old version because the freshest one is not SourceLink-ed.

vasily-kirichenko:
Thanks a lot, it's great!

dungpa:
@forki How did you do migration? Modify `dependencies` file and run `paket install`? It's not entirely clear from Paket docs http://fsprojects.github.io/Paket/groups.html

forki:
yes. install tries to detect what changed in the deps file and comes up with minimal changes in lock file.

dungpa:
Got it. Thanks.

PR #1161, opened 2015-10-13
outline nested bindings, matches and CEs

vasily-kirichenko:
improve outlining tests


dungpa:
Thinking about this a bit more, I agree with @7sharp9's rationale. Perhaps we should restrict outlining to top-level constructs because (1) we don't often need fine-grained outlining (2) AST traversal would be much slower. 

What do you think?

tests/FSharpVSPowerTools.Core.Tests/OutliningTests.fs
```diff
@@ -38,128 +41,133 @@ let (=>) (source: string) (lineRanges: (int * int) list) =
             let multilineRanges =
                 getOutliningRanges tree
                 |> Seq.filter (fun r -> r.StartLine <> r.EndLine)
-                |> Seq.map (fun r -> r.StartLine, r.EndLine)
+                |> Seq.map (fun r -> r.StartLine, r.StartColumn, r.EndLine, r.EndColumn)
                 |> List.ofSeq
-            CollectionAssert.AreEquivalent (List.sort lineRanges, List.sort multilineRanges)
+            CollectionAssert.AreEquivalent (List.sort ranges, List.sort multilineRanges)
```
dungpa: I would just name them `expectedRanges` and `actualRanges` to avoid confusion.

`CollectionAssert` members don't give good error messages on failing tests. I think our `assertEqual` could do better jobs here.

vasily-kirichenko: Done.

vasily-kirichenko:
@dungpa I disagree. It's desirable to collapse nested functions, large PMs and if...else and object expressions and so on.

7sharp9:
Non top level things like nested function tend to get poor symbol as they are treated differently to top level stuff so Im always wary of those.  Im not sure what they look like in the parse tree though.

7sharp9:
I would also tend to think loads of nested functions means your pretty close to refactoring out to internal modules too :-)

dungpa:
@vasily-kirichenko Ok, my concern is excessive outlining that we don't know where to stop. Let's see how it goes.

PR #1164, opened 2015-10-13
Add F# Lint's underline greedier.

yukitos:
In particular:
1. When warning's range is in a line (i.e. StartLine and EndLine is the same), entire text which has some warnings will be marked.
2. When it spreads over multiple lines, text which is in the StartLine and is from StartColumn to the end of the StartLine will be marked.

### DIFF
--- a/src/FSharpVSPowerTools.Logic/LintTagger.fs
+++ b/src/FSharpVSPowerTools.Logic/LintTagger.fs
@@ -69,8 +69,10 @@ type LintTagger(textDocument: ITextDocument,
                             let r = warn.Range
                             let endCol =
                                 if r.StartLine = r.EndLine then
-                                    min r.EndColumn (r.StartColumn + 2)
-                                else r.StartColumn + 2
+                                    r.EndColumn
+                                else
+                                    let line = buffer.CurrentSnapshot.GetLineFromLineNumber(r.StartLine - 1).GetText()
+                                    line.Length
                             let range =
                                 Range.mkRange "" 
                                     (Range.mkPos r.StartLine r.StartColumn)


yukitos:
For reference:
image

vasily-kirichenko:
It seems many people complain about too short squiggles. I'm ok with this change. @dungpa @duckmatt what do you think?

dungpa:
I'm fine with the change. Please merge.

duckmatt:
Looks good

vasily-kirichenko:
@yukitos thanks!

yukitos:
My great pleasure!

vasily-kirichenko:
Mm. I think it's too obtrusive on long lines, like this:

image

yukitos:
It seems that FSharpLint reports multiline warning about the processTask function.
Do we need to reduce the underline until the first whitespace characters?

vasily-kirichenko:
@yukitos maybe. It's hard to decide without looking at what it looks like.

yukitos:
OK. I'll try to change the behavior.

yukitos:
Here is my first trail result:
image

As this image, a lot of underlined "fun" keyword can be found in FSharpVSPowerTools.v14.sln ;)
Does this look fine?

vasily-kirichenko:
Now it's too short ;)

What about cutting the line by, say, 20 chars maximum, but keep words in their entirety? Like this:

`too long line very too long line line line` => `too long line very`

dungpa:
It is potentially confusing with truncating until the first space. If the range starts with `((<space>`, we will see underlying warnings on `((`, which is weird.

I think it's OK with current behaviour. I don't mind Lint tools being a bit intrusive. It will force people either fix the warning or disable the relevant warning, which is its main purpose :-).

vasily-kirichenko:
Ok, let it be as it is (full first line is highlighted).

yukitos:
Then I'll leave the code as it is :+1:

PR #1167, opened 2015-10-14
Automatic upload to Visual Studio gallery

forki:
This is taken from Paket.VisualStudio project. I was not able to test it here since I don't own your gallery account.

In oder to use this you need a gallerycredentials.txt in the root. First line your email, second line password. The file is already gitignored.

This is how it looks in Paket.VisualStudio:

gallery


dungpa:
:+1:

PR #1168, opened 2015-10-14
do not parse and check the whole project after each `ParseAndCheckFileInProject` call

vasily-kirichenko:
+ calculate unused opens in Fast SyntaxConstructClassifier's state

It perform well having the following features on:

image

The only feature that requires full project checking after any change in a file is "Gray out unused declarations". So on large *projects* it must be turn off.

### DIFF
--- a/src/FSharpVSPowerTools.Core/LanguageService.fs
+++ b/src/FSharpVSPowerTools.Core/LanguageService.fs
@@ -210,9 +210,7 @@ type LanguageService (?fileSystem: IFileSystem) =
                    while files.ContainsKey filePath &&
                          (not (files.TryUpdate (filePath, BeingChecked, Checked)
                                || files.TryUpdate (filePath, BeingChecked, NeedChecking))) do
-                       let state = files.[filePath]
-                       debug "[LanguageService] %s is in state = %A, waiting for Checked or NeedChecking..." filePath state
-                       do! Async.Sleep 200
+                       do! Async.Sleep 20
                    
                    debug "[LanguageService] Change state for %s to `BeingChecked`" filePath
                    debug "[LanguageService] Parse and typecheck source..."
@@ -354,10 +352,7 @@ type LanguageService (?fileSystem: IFileSystem) =
           | Some results -> return results
           | None -> 
               debug "Parsing: Trigger parse (fileName=%s)" fileName
-              let! results = parseAndCheckFileInProject(fileName, src, opts)
-              debug "Worker: Starting background compilations"
-              checkerInstance.StartBackgroundCompile opts
-              return results
+              return! parseAndCheckFileInProject(fileName, src, opts)
       }
 
   /// Get all the uses of a symbol in the given file (using 'source' as the source for the file)
--- a/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
+++ b/src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
@@ -137,7 +137,7 @@ type SyntaxConstructClassifier
 
     let getOpenDeclarations source project ast getTextLineOneBased (pf: Profiler) = async {
         let! entities = pf.TimeAsync "GetAllEntities" <| fun _ ->
-            vsLanguageService.GetAllEntities(textDocument.FilePath, source, project)    
+            vsLanguageService.GetAllEntities(textDocument.FilePath, source, project)
 
         return! pf.TimeAsync "getOpenDeclarations" <| fun _ -> async {
             let qualifyOpenDeclarations line endCol idents = async { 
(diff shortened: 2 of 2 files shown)


dungpa:
Why calculating unused opens in Fast SyntaxConstructClassifier's state? Is there any performance implication?

vasily-kirichenko:
It does not need full project(s) checking and it's still called asynchronously. Such a decision allows turn it on and get nice performance. Actually the only feature which require project rechecking is Gray out unused declarations, so user can now switch it off and get good perf.

dungpa:
I see. :+1:

dsyme:
The StartBackgroundCompile call is a heuristic: sometimes good, sometimes overly optimistic. The heuristic is assuming that there's enough CPU and memory to go through and background compile the current project. It does this so that everything in the project analysis is hot and ready should the user switch to another file, or ask for a rename refactoring etc.  The Visual F# Tools make a similar choice.  

This assumption seems valid when memory is low and CPU is not stressed.  It likely makes for a better experience in small-medium projects (< 50,000 lines say).  But it's certainly just a heuristic.

So I suspect the right thing here is to make this call in some kind of "low stress" mode. That would need a good heuristic to detect that though.  I would say "memory < 1.5 GB, multiple cores available, not in battery-save mode, project size < 50,000 lines".  

cheers
don

vasily-kirichenko:
How starting the background compilation can be any good in case where the user changes a single file continuously for long time (which is a very common use case)?

I can hardly see in what circumstances this logic could be useful at all.

dsyme:
In the presence of multiple edits, the background compile won't run to completion on each StartBackgroundCompile request - on the first edit it will start it's background work, and consume one background time slice (~50ms, see https://fsharp.github.io/FSharp.Compiler.Service/queue.html) and then start again (because more edits have happened)

The heuristic is good in some situations: for example, if you stop editing then the CPU is being used productively. All IDEs have some kind of background compilation like this.

If you really are editing all the time then it's not useful.  

So yes, if you can successfully detect the difference then only call StartBackgroundCompile when the user has stopped editing for a bit.  Obviously it's a game between you and the user though...  

FWIW Roslyn has heuristics like "disable GC for 5 seconds after the user has typed a character" which are a similar game.  If you type really really fast all the time then you will never get a GC....

src/FSharpVSPowerTools.Core/LanguageService.fs
```diff
@@ -354,10 +352,7 @@ type LanguageService (?fileSystem: IFileSystem) =
           | Some results -> return results
           | None -> 
               debug "Parsing: Trigger parse (fileName=%s)" fileName
-              let! results = parseAndCheckFileInProject(fileName, src, opts)
-              debug "Worker: Starting background compilations"
-              checkerInstance.StartBackgroundCompile opts
```
dsyme: @vasily-kirichenko I'm not sure we should be removing this until we've got to the of CPU and memory usage issues.  

VFT uses a background compile - for example, the background compile for a project is kicked off whenever you switch to a file in that project.  This of course uses CPU when editing and memory to store the results - but it means that lots of requests (e.g. intellisense and symbols) are very quick. 

In principle it's reasonable for VFPT to be doing the same - if you don't, then you will have to do more compute when doing things like rename refactoring .  The question I'd relaly like to get to the bottom of before making fairly large behavioural changes like this is why adding the VFPT/FCS background build uses so much memory, and also whether particular VFPT/FCS features are causing unnecessary CPU load.

So I'd recommend holding off on this change just at the moment.

dsyme: @vasily-kirichenko I'm not sure we should be removing this until we've got to the of CPU and memory usage issues.  

VFT uses a background compile - for example, the background compile for a project is kicked off whenever you switch to a file in that project.  This of course uses CPU when editing and memory to store the results - but it means that lots of requests (e.g. intellisense and symbols) are very quick. 

In principle it's reasonable for VFPT to be doing the same - if you don't, then you will have to do more compute when doing things like rename refactoring .  The question I'd relaly like to get to the bottom of before making fairly large behavioural changes like this is why adding the VFPT/FCS background build uses so much memory, and also whether particular VFPT/FCS features are causing unnecessary CPU load.

So I'd recommend holding off on this change just at the moment.

vasily-kirichenko:
@dsyme I removed the background compilation to fix your issue when after each file change VS was consuming CPU for 30 seconds. I've used this build for 2 days now and I must to say that editing experience is **tremendously** better than before, especially on large complex project (especially if TPs are used). It's so cool that I don't think we should turn the old behavior on again. 

I have a question though. Does the following reflect how things actually work?

* load a project -> nothing happens
* open a file -> call `ParseAndCheckFileInProject` -> FCS actually checks the whole project
* edit the file -> call `ParseAndCheckFileInProject` -> FCS checks the file (only)
* **are check results cached by FCS for the rest of files in the project are outdated now**? Should we call `ParseProject` / `StartBackgroundCompilation` in order to Rename refactoring / Find all refs work properly?

PR #1171, opened 2015-10-15
Outlining

cloudRoutine:
If Then Else clauses collapse
For loops and ForEach loops collapse
Hint text restricted to the range being collapsed
Abstract members collapse
`...` placed at the end of first line, no longer eats the `=`


vasily-kirichenko:
Could you flatten the commits, I mean removing the merges? Rebase interactive should work.

cloudRoutine:
rebase failure, something's wrong with my git config that I haven't been able to figure out

PR #1177, opened 2015-10-16
DepthTagger retrieves project checker options from ProjectProvider instead of getting it from FSharpChecker

vasily-kirichenko:


vasily-kirichenko:
Also I revern `QuickInfoMargin` delay to 200 ms since I don't think it has anything with VFPT performance and 1000 ms is too long, it's refreshed with annoying latency.

src/FSharpVSPowerTools.Logic/DepthTagger.fs
```diff
@@ -30,10 +39,15 @@ type DepthTagger(buffer: ITextBuffer, filename: string, languageService: VSLangu
...
-        async { 
+        asyncMaybe { 
             let snapshot = buffer.CurrentSnapshot // this is the possibly-out-of-date snapshot everyone here works with
-            let source = snapshot.GetText() 
-            let! ranges = languageService.CheckerAsync (DepthParser.getNonoverlappingDepthRanges source filename)
+            let source = snapshot.GetText()
+            let dte = serviceProvider.GetService<EnvDTE.DTE, SDTE>()
+            let! doc = dte.GetCurrentDocument doc.FilePath
+            let! project = projectFactory.CreateForDocument buffer doc
+            let! opts = languageService.GetProjectCheckerOptions project |> liftAsync
```
dsyme: Are the results of this saved?  Or are you recomputing the options on each change of the document? Just checking.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -136,7 +136,7 @@ type QuickInfoMargin (textDocument: ITextDocument,
             |> Async.StartInThreadPoolSafe
         | None -> updateQuickInfo (None, None)
 
-    let docEventListener = new DocumentEventListener ([ViewChange.layoutEvent view; ViewChange.caretEvent view], 1000us, updateAtCaretPosition)
+    let docEventListener = new DocumentEventListener ([ViewChange.layoutEvent view; ViewChange.caretEvent view], 200us, updateAtCaretPosition)
```
dsyme: What's this change? thx

src/FSharpVSPowerTools.Logic/DepthTagger.fs
vasily-kirichenko: Yes, `ProjectFactory` is actually a cache of project options. This is what we use in all other features.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
dsyme: OK, I see the comment in the PR.  Does this increase the frequency of doing updates to the QuickInfo margin?  will that result in more FCS requests?  (Currently I'm seeing cases where the FCS request queue grows to be very long very quickly - and the requests aren't yet being cancelled when new requests come in - so changes that increase the number of FCS requests have to be done with care)

vasily-kirichenko: Shouldn't https://github.com/fsharp/FSharp.Compiler.Service/pull/445 fix it so we can now don't worry about number of requests? 

Many of our features use 200..500 ms delay.

src/FSharpVSPowerTools.Core/DepthParser.fs
```diff
@@ -361,7 +361,7 @@ module DepthParser =
     /// Get non-overlapping ranges, where each range spans at most a single line, and has info about its "semantic depth".
     /// Note: The 'filename' is only used e.g. to look at the filename extension (e.g. ".fs" versus ".fsi"), this does not try to load the file off disk.  
     ///       Instead, 'sourceCodeOfTheFile' should contain the entire file as a giant string.
-    let getNonoverlappingDepthRanges sourceLines filename (checker: FSharpChecker) =
+    let getNonoverlappingDepthRanges (sourceLines, filename, project, checker: FSharpChecker) =
```
dungpa: The `project` parameter is not used. We should pass it into `getRanges` in order to remove `checker.GetProjectOptionsFromScript` call.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
vasily-kirichenko: With 1s delay it feels like a slow feature:

dungpa:
I think similar refactoring should be done for `XmlDocParser` as well.

vasily-kirichenko:
all done, thanks

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
dsyme: Ah, I see.  I'm mostly concerned about flooding FCS when there are lots of rapid edits.  FCS will cache one ParseAndCheckFileInProject request for each file, keyed on the source text.  If the source text is changing, this cache is naturally of no use.  So asyncs that flood FCS with requests with different source texts will quickly create too much work.

One way to help here is to cancel outstanding asyncs relating to a previous source text once an edit has been made.  Another is to put in a delay before starting the request (the delay is in itself a form of cancellation - it's giving a pause to allow a cancellation to come through).  The next FCS nuget will include the ability to cancel requests, see [this](https://github.com/fsharp/FSharp.Compiler.Service/pull/445/files#diff-d784b269182ed2ba106aaa72b68ef9dfR48)

PR #1178, opened 2015-10-16
Update to FCS 1.4.2.0

dsyme:


dsyme:
This exposed a bug in the newly released FCS 1.4.1: https://github.com/fsharp/FSharp.Compiler.Service/pull/447

vasily-kirichenko:
Updated to 1.4.2. Waiting for CI.

dsyme:
I have been using this for two sessions each over half an hour on the FSharp.Compiler.Service project.  Previously this was impossible with VFPT enabled at all for any "active" features.

It feels like things have really improved - memory is still large (as expected) but I got a lot of work done staying in the 1.3GB-1.6GB range.  The UI was very responsive the whole time and all the VFPT features felt smoother.  I even started to turn features back on and was able to run with unused namespace gray-out for the second session.

From that experience I'd say this is ready to go in!  It feels like we've made some progress at last...

dsyme:
If people want to try this, here it is: https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/2.2.0-a1510161958/artifacts

vasily-kirichenko:
@dsyme Fantastic! I cannot believe you fixed the memory issue.

vasily-kirichenko:
@dsyme I tried to set `keepAssemblyContents = false` and it behaved even better. Is it worth to set this flag or there are some hidden cost behind it?

dsyme:
That should default to false: https://github.com/fsharp/FSharp.Compiler.Service/blob/03e3d0260fdd670999b338b8f2cd8df8b1411899/src/fsharp/vs/service.fs#L3048

You need to set this if you want the typed expression tree (AssemblyContents) - and yes, it will use more memory.

PR #1179, opened 2015-10-17
FCS project cache size reduced from 50 to 10

vasily-kirichenko:
also set `FSharpChecker`'s `keepAssemblyContents = false`

### DIFF
--- a/src/FSharpVSPowerTools.Core/LanguageService.fs
+++ b/src/FSharpVSPowerTools.Core/LanguageService.fs
@@ -166,11 +166,12 @@ type LanguageService (?fileSystem: IFileSystem) =
   let handleCriticalErrors e file source opts = 
       errorHandler |> Option.iter (fun handle -> handle e file source opts)
 
-  // Create an instance of interactive checker. The callback is called by the F# compiler service
-  // when its view of the prior-typechecking-state of the start of a file has changed, for example
-  // when the background typechecker has "caught up" after some other file has been changed, 
-  // and its time to re-typecheck the current file.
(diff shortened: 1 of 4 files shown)


vasily-kirichenko:
It should prevent uncontrolled memory usage on many-project solutions.

src/FSharpVSPowerTools.Core/LanguageService.fs
```diff
@@ -166,11 +166,12 @@ type LanguageService (?fileSystem: IFileSystem) =
...
 
-  // Create an instance of interactive checker. The callback is called by the F# compiler service
-  // when its view of the prior-typechecking-state of the start of a file has changed, for example
-  // when the background typechecker has "caught up" after some other file has been changed, 
-  // and its time to re-typecheck the current file.
-  let checkerInstance = FSharpChecker.Create (projectCacheSize = 50, keepAllBackgroundResolutions = false)
+  // Create an instance of interactive checker.
+  let checkerInstance = 
+    FSharpChecker.Create(
+        projectCacheSize = 10,
```
dsyme: Can you make this an option in the options panel?

Or we could change the max-memory behaviour of FCS to reduce this by 10% each time the max-memory condition is hit.

vasily-kirichenko: I like the latter idea very much. It would be cool if FCS notify about this via an event (we could write it into Output window).

vasily-kirichenko:
I'm merging it for now. When the dynamic caches size shrinking is available, we will replace this logic with it.

PR #1182, opened 2015-10-18
Fix - Tooltips contain extra end line characters

cloudRoutine:
The issue was caused by only accounting for lines that terminate with a
`\n`, as per the configuration of the VFPT repo.  Now the string
extracted from the span of the buffer being outlined will be split
accounting for the linebreak characters of windows, linux, and osx.

- added type extension to `String` for replace
- added type extension to `String` for splitting into a string array
accounting for linux, windows, osx
- added extension method ^ same functionality


vasily-kirichenko:
Great.

PR #1184, opened 2015-10-18
FSharpChecker.ImplicitlyStartBackgroundWork <- false

vasily-kirichenko:
With the flag set to `true` editing files is not smooth. For example, editing `VSUtils.fs`, which is at top of the largest VFPT project, is not comfortable enough. 

Maybe we should add a setting for this.

### DIFF
--- a/src/FSharpVSPowerTools.Core/LanguageService.fs
+++ b/src/FSharpVSPowerTools.Core/LanguageService.fs
@@ -173,6 +173,8 @@ type LanguageService (?fileSystem: IFileSystem) =
         keepAllBackgroundResolutions = false,
         keepAssemblyContents = false)
 
+  do checkerInstance.ImplicitlyStartBackgroundWork <- false
+
   let checkerAsync (f: FSharpChecker -> Async<'a>) = 
     let ctx = System.Threading.SynchronizationContext.Current
     async {


vasily-kirichenko:
I suspect we need to run `CheckAndParseProject` when Rename refactoring and Find all refs are launched. Also we should run `ParseProject` (is there this one?) before Navigate to in order to reparse/recheck all files in current project.

vasily-kirichenko:
I've tested it and it seems to work ok. 

I still don't think that starting background compilation on every change is a good thing. In theory it performs in a separate thread and does not affect UI experience in any way, but currently it's not the case. Having background compilation on causes lags in typing, which users don't like very much.

dsyme:
Please don't set this to false as yet (though by all means have an option to do it manually, with the default to true).  The experience on my 4-core laptop was very good, even for a large project.   And for a small project it is entirely non-intrusive. If you turn this off, then the first time you do an operation like find-all-references when browsing a large project, you will get a full project compile when "Initializing..." shows.  You'll also get that after each edit. That's not a good experience.

I'd really like to get the defaults right in FCS, as tools like Xamarin, Emacs, Atom will all inherit much the same behaviour.   I've found the latest VFPT is already 10x better on responsiveness, so let's take it one step at a time?

The Visual F# Tools do a background compile - as do all other IDE tools - so it is reasonable for you to do one as well.  The current FCS you're using has a delay of 1 second of idleness before starting work - perhaps increase that to 2 or 3 seconds  Note the background compile is entirely on a background CPU thread and stops pretty much as soon as any FCS request comes through.

My recommendations:

1. leave the defaults the way they are.  

2. add project settings for ImplicitlyStartBackgroundWork  and PauseBeforeBackgroundWork 

3. If you want to relieve CPU pressure further by default, then first turn off FSharpLint by default, which takes ~50% of your budget for CPU.

vasily-kirichenko:
Thanks, Don. OK, leave it as it is. The memory leaks makes this irrelevant anyway.

dsyme:
@vasily-kirichenko Do we still have known memory leaks?  Thanks.

dsyme:
@vasily-kirichenko I tried this scenario with the latest VFPT from CI:

1. open FSharpVSPowerTools.sln 
2. edit at the top of VSUtils.fs (Colorization on, QuickInfoPanel on + FSHarpLint off) 
3. do various re-entry of existing code and refactor-renamings for about 3 minutes

Results on my 4-core laptop:

- no UI blocks 
- memory stayed < 1GB
- a couple of slow updates to VFPT colorization and QuickInfoPanel that would be worth investigating (is the FCS queue getting too long?)

One next area to improve is to cancel the async's arising from QuickInfoPanel and colorization once a subsequent request ordered.

Best 
Don

vasily-kirichenko:
Don, I've reached 2GB working in VFPT yesterday. I didn't do anything unusual. On the other hand, I've been working in my large solution at work (93 F# projects + 80 C# ones) for the whole day and VS's stayed around 1.6GB reaching 2.5GB sometimes. Editing has been smooth. 

Then I executed Find all references on a symbol which required to type check all 93 F# projects. VS occupied 2.5GB after 20 or so projects after which it became unusable and I had to restart it.

Checker's project cache is 10 now, so it should be possible to type check any number of projects in roughly constant memory.

PR #1191, opened 2015-10-18
Fix NullReferenceException

bleis-tift:
I found missing patterns.
I don't know why FsCheck doesn't generate (null, notNull) pattern and I don't know well about FsCheck.
So this fix is dirty...

### DIFF
--- a/src/FSharpVSPowerTools.Core/Utils.fs
+++ b/src/FSharpVSPowerTools.Core/Utils.fs
@@ -103,6 +103,7 @@ module Array =
         match x, y with
         | null, null -> true
         | [||], [||] -> true
+        | null, _ | _, null -> false
         | _ when x.Length <> y.Length -> false
         | _ ->
             let mutable break' = false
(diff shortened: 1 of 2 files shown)


bleis-tift:
Current implementation of FsCheck applies `Array.ofList` to the `list` that generated by `Gen.listOfLength` to generate array.

https://github.com/fscheck/FsCheck/blob/17c2c6de1f009f27758540eb3feb63010c88cddb/src/FsCheck/Arbitrary.fs#L498
https://github.com/fscheck/FsCheck/blob/17c2c6de1f009f27758540eb3feb63010c88cddb/src/FsCheck/Gen.fs#L401
https://github.com/fscheck/FsCheck/blob/17c2c6de1f009f27758540eb3feb63010c88cddb/src/FsCheck/Gen.fs#L393

`Gen.listOfLength` doesn't generate `null`.
So FsCheck doesn't generate `null` for array.

vasily-kirichenko:
Great catch, thanks!

PR #1193, opened 2015-10-20
add a failing test exposing wrong enum cases colorization as mutable vars

vasily-kirichenko:


vasily-kirichenko:
Fixed. The bug in FCS is still should be fixed.

PR #1198, opened 2015-10-21
Updated to reflect bitbuckets url change

TWith2Sugars:


vasily-kirichenko:
@rojepp I know you use BitBucket a lot, it would be great if you test this change.

TWith2Sugars:
Bitbucket is a bit flaky today, took a few tries to test this my self.

-----Original Message-----

rojepp:
@TWith2Sugars How did you test this? You'd need to use a lib that lives at bitbucket and has source links, right?

src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
```diff
@@ -355,7 +355,8 @@ type GoToDefinitionFilter(textDocument: ITextDocument,
                         | String.StartsWith "https://raw.github.com" _-> formattedGithubUrl
                         | String.StartsWith "https://github.com" _-> formattedGithubUrl
                         | String.StartsWith "https://bitbucket.org" _-> 
-                            sprintf "%s#cl-%d" (url |> replace "/raw/" "/src/") r.StartLine
+                            let fileName = Path.GetFileName r.FileName
+                            sprintf "%s?fileviewer=file-view-default#%s-%d" (url |> replace "/raw/" "/src/") fileName r.StartLine
```
rojepp: Bitbucket source urls with line number looks like this: 
> https://bitbucket.org/IntelliFactory/websharper.google.maps/src/45a6136882edfecd168e0c43c17eced387890be4/build.cmd?at=default&fileviewer=file-view-default#build.cmd-7

This looks like what the code does to me. Except the branch is missing (`?at=default`). I haven't tested the code yet.

TWith2Sugars:
Yeah the company has private repos with sourcelink packages. Just tested with that.

-----Original Message-----

vasily-kirichenko:
@TWith2Sugars thanks
@rojepp I see you have not found any source linked repos :)

src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
TWith2Sugars: Doesn't need the branch as has the commit hash.

rojepp:
I think this is ready to merge.

TWith2Sugars:
One thing I don't get, why do we perform a http request to check if the url exists?

Wouldn't removing that and just letting the browser try to display it be better? If

vasily-kirichenko:
@TWith2Sugars maybe we check the URL to give better feedback?

TWith2Sugars:
All it does is prevent the browser from opening it, if the 3rd part host is down it feels like this feature isn't working. I've noticed this with bitbucket today. There is a little info status bar but it's easily missed.

I've submitted another pr to remove this is if people think it's right: https://github.com/fsprojects/VisualFSharpPowerTools/pull/1199

PR #1199, opened 2015-10-21
removed the http check in navigateTosource

TWith2Sugars:
Since it doesn't seem to do anything other than a check I removed the use of the http client.

Everything works the same still just the browser pops up a little faster.

If we are to keep this then perhaps we need to change how we handle the HttpClient since the consensus is the create 1 instance and reuse it: http://stackoverflow.com/a/15708633/35389

### DIFF
--- a/src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
+++ b/src/FSharpVSPowerTools.Logic/GoToDefinitionFilter.fs
@@ -327,47 +327,38 @@ type GoToDefinitionFilter(textDocument: ITextDocument,
             if String.IsNullOrEmpty s then None else Some s)
 
     let navigateToSource (fsSymbol: FSharpSymbol) (url: string) = 
-        async {
-            use handler = new HttpClientHandler(UseDefaultCredentials = true)
-            use http = new HttpClient(handler)
-            let! response = http.GetAsync(url) |> Async.AwaitTask
-            if response.IsSuccessStatusCode then
-                fsSymbol.ImplementationLocation
-                |> Option.iter (fun r ->
-                    // NOTE: other source code hostings might have different url templates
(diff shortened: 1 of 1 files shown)


vasily-kirichenko:
I don't mind to remove the additional check. 
@dungpa what do you think?

dungpa:
Sounds like a good idea.

src/FSharpVSPowerTools.Logic/FsiReferenceCommand.fs
```diff
@@ -174,32 +174,25 @@ type FsiReferenceCommand(dte2: DTE2, mcs: OleMenuCommandService) =
...
-            let! project = Option.ofNull project
-            let! currentConfiguration = 
-                try Some (project.ConfigurationManager.ActiveConfiguration.ConfigurationName.ToLower())
-                with e -> 
-                    Logging.logError "[FsiReference] Cannot get current configuration name: %O" e
-                    None
-            let loadRefsFileName = sprintf "load-references-%s.fsx" currentConfiguration
+        Option.ofNull project
+        |> Option.iter (fun project ->
+            let loadRefsFileName = "load-references.fsx"
```
bleis-tift: @vasily-kirichenko  Are these changes really OK?
These changes look like to revert a3d828d170063ec6612951151dc53c7d2a794266 .

vasily-kirichenko: Yes, it's. I'm just tired to commit the generated files each time I switch configuration and build.

TWith2Sugars: Whoops, didn't realise I commited this.

vasily-kirichenko: Np, I cherry picked my commit and push it into master.

PR #1200, opened 2015-10-21
Add colored outlining hints

dungpa:
It's similar to Roslyn's outlining hints in VS 2015.

image


vasily-kirichenko:
Have not tested yet, but it looks *awesome*!

vasily-kirichenko:
Could you rebase?

dungpa:
It doesn't conflict with anything. Why should we rebase?

PR #1201, opened 2015-10-22
revert FCS project cache size to 50 (was 10)

vasily-kirichenko:
Unfortunately, with project size set to 10 it turned out that projects which reference large number of other projects in same solution, are not processed *at all*. Symptoms:

* One of CPU cores is load at 100% *infinitely*
* devenv.exe private set increases linearly until reaches 2.5GB in about 1 minute
* VS becomes unusable due to GC

After reverting the cache size to 50 everything is ok.


dsyme:
Could you enter an FCS bug on this? It feels like project results need to be held more strongly while the project checking is performed. Thanks

PR #1207, opened 2015-10-24
add "BackgroundCompilation" setting

vasily-kirichenko:
Sorry guys, but background compilation slows down VS too much.


dungpa:
Cool. While you're at it, please add 'Project Cache Size' option as well.

vasily-kirichenko:
Great idea, done.

dungpa:
:+1:

PR #1209, opened 2015-10-24
Make logging more lightweight

dungpa:
Avoid the cost of building printf specifiers although there is no printing.

See http://stackoverflow.com/questions/31442608/how-to-wrap-sprintf-conditionally-in-f.

I was trying to keep the printf-like logging functions but we can't get zero-cost abstraction here.


vasily-kirichenko:
Great solution, simple and transparent.

PR #1210, opened 2015-10-24
Remove Strict mode option

dungpa:
It was a desperate effort to bypass a serious bug in FCS. The bug has been fixed long ago; it's safe to remove the hack.


vasily-kirichenko:
Great, the less code the better :)

PR #1212, opened 2015-10-24
Treat warnings as errors

dungpa:
For all F# projects:
- Raise to warning level 5
- Enable unused value warning
- Treat warnings as errors

It's a bit strict, but it will make maintenance easier in the long term.


vasily-kirichenko:
Great.

PR #1218, opened 2015-10-28
Build script also updates the vsix manifest version number.

TWith2Sugars:


build.fsx
```diff
@@ -67,8 +69,8 @@ Target "AssemblyInfo" (fun _ ->
   let shared =
       [ Attribute.Product project
         Attribute.Description summary
-        Attribute.Version release.AssemblyVersion
-        Attribute.FileVersion release.AssemblyVersion ] 
+        Attribute.Version version
```
dungpa: I think it's better to keep `release.AssemblyVersion` here. Assembly version doesn't have any effect in installing the extension.

build.fsx
```diff
@@ -86,6 +88,17 @@ Target "AssemblyInfo" (fun _ ->
       (Attribute.InternalsVisibleTo "FSharpVSPowerTools.Tests" :: Attribute.Title "FSharpVSPowerTools.Logic.VS2015" :: shared) 
 )
 
+Target "VsixManifest" (fun _ ->
+    let manifest = "./src/FSharpVSPowerTools/source.extension.vsixmanifest"
+    let doc = new XmlDocument() in
```
dungpa: Maybe we should use `XmlDocument(PreserveWhitespace=true)`?

build.fsx
```diff
@@ -301,6 +314,7 @@ Target "All" DoNothing
 "Clean"
   =?> ("BuildVersion", isAppVeyorBuild)
   ==> "AssemblyInfo"
+  ==> "VsixManifest"
```
dungpa: We should only run this on AppVeyorCI. Locally, we don't want FAKE build to override our vsix version.

I think a better way is to extract Version value (e.g. 2.1.3) and append build numbers into it?

TWith2Sugars:
Hows that?

dungpa:
One last thing:

```
node.Attributes.GetNamedItem("Version").Value <- version
```

should be

```
let currentVersion = node.Attributes.GetNamedItem("Version").Value
node.Attributes.GetNamedItem("Version").Value <- sprintf "%s.%s" currentVersion AppVeyor.AppVeyorEnvironment.BuildNumber
```

As it is, users have to uninstall builds from AppVeyor when the official one is released.

TWith2Sugars:
Would they? The build number for the released on would also be incremented.

Either way its done.

dungpa:
@FeodorFitsner Is there any bug on AppVeyor deployment? We see builds from other users and some pushes are not built at all.

FeodorFitsner:
Looking in that. Can you drop a link to a build please?



- Feodor

TWith2Sugars:
https://ci.appveyor.com/project/dungpa/visualfsharppowertools/history

TWith2Sugars:
and https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/1.0.1753

FeodorFitsner:
Cancel queued builds and start new one - it should be fine now. Please let me know about the results.

dungpa:
@FeodorFitsner Everything is ok now.

dungpa:
@TWith2Sugars Great, thanks.

PR #1219, opened 2015-10-28
Fix loosing coloring

vasily-kirichenko:
WIP


vasily-kirichenko:
May be tested. Seems to work.

dungpa:
Great.

PR #1221, opened 2015-10-30
Utils

cloudRoutine:
utility functions and minor performance enhancements


src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -35,28 +60,44 @@ module Array =
         | null -> nullArg argName 
         | _ -> ()
 
+    /// Optimized arrays equality. ~100x faster than `array1 = array2` on strings.
+    /// ~2x faster for floats
+    /// ~0.80% for ints
```
vasily-kirichenko: 0.8%? Or 80%? Or 0.8x faster?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -581,6 +630,45 @@ module String =
...
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module StringBuilder =
+    /// Pipelining function for appending a string to a stringbuilder
+    let inline append   (str:string) (sb:StringBuilder) = sb.Append str
+
+    /// Pipelining function for appending a string with a '\n' to a stringbuilder
+    let inline appendln (str:string) (sb:StringBuilder) = sb.AppendLine str
+    
+    /// SideEffecting function for appending a string to a stringbuilder
+    let appendi (str:string) (sb:StringBuilder) = sb.Append str |> ignore
```
vasily-kirichenko: Why not `inline` this one?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -581,6 +630,45 @@ module String =
...
+
+    /// SideEffecting function for appending a string with a '\n' to a stringbuilder
+    let appendlni (str:string) (sb:StringBuilder) = sb.AppendLine str |> ignore
```
vasily-kirichenko: Same as above.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -37,51 +39,50 @@ type QuickInfoMargin (textDocument: ITextDocument,
...
+                errors |> Option.map (fun errors ->
+                    (StringBuilder (), errors) ||> Array.fold (fun sb (severity, err) -> 
+                        let errorls = List.map String.trim err
+                        let title = 
+                            match errorls with
+                            | [_] -> sprintf "%s" <| string severity
+                            | _ -> sprintf "%s (%d)" (string severity) errorls.Length
+                        (sb |> append title |> append ": " |> errorString errorls |> append " ")) |> string)  
+                // show type info if there aren't any errors
+                |>  Option.orElse (tooltip |> Option.bind (fun tooltip ->
```
vasily-kirichenko: I think the first  argument passed to `Option.orElse` (`tooltip |> ...`) is always evaluated, even if the second argument is `Some ...`. We need something like `Option.orElseWith: (unit -> 'b) -> 'a option -> 'b` to make it lazy.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -37,51 +39,50 @@ type QuickInfoMargin (textDocument: ITextDocument,
...
                     else str)))
-            |> Option.getOrElse ""
+                // if there are no results the panel will be empty
+                |> Option.getOrElse ""
+            model.QuickInfo <- currentInfo
+        lock updateLock updateFunc      
+        
+    // helper function in the form required by mapNonEmptyLines
+    let flattener (sb:StringBuilder) (str:string) : StringBuilder =
+            if str.Length > 0 && Char.IsUpper str.[0] then (sb.Append ". ").Append (String.trim str)
```
vasily-kirichenko: `Array.tryHead` + PM?

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -37,51 +39,50 @@ type QuickInfoMargin (textDocument: ITextDocument,
...
+    // helper function in the form required by mapNonEmptyLines
+    let flattener (sb:StringBuilder) (str:string) : StringBuilder =
+            if str.Length > 0 && Char.IsUpper str.[0] then (sb.Append ". ").Append (String.trim str) 
+            else (sb.Append " ").Append (String.trim str)
+                
 
-    let flattenLines (x: string) =
+    let flattenLines (x: string) : string =
+        let appendDot (x:string) = 
+            if x.Length > 0 && x.[x.Length - 1] <> '.' then x + "." else x
```
vasily-kirichenko: `Array.tryLast`?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -517,28 +561,33 @@ module String =
         match str with
         | null -> null
         | str ->
-            match str.ToCharArray() |> Array.toList with
-            | [] -> str
-            | h :: t when Char.IsUpper h -> String (Char.ToLower h :: t |> List.toArray)
+            match str.ToCharArray () with
+            | [||] -> str
+            | arr when Char.IsUpper arr.[0] ->
```
vasily-kirichenko: `Array.tryHead` + PM with a AP `UpperChar`? (see same stuff below)

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -517,28 +561,33 @@ module String =
...
+                let mutable darr = chars |> Array.takeWhile Char.IsDigit
+                Array.revInPlace &darr; darr
+            String digits
             |> function
                | "" -> str, None
                | index -> str.Substring (0, str.Length - index.Length), Some (int index)
 
+    /// Remove all trailing and leading whitespace from the string
+    /// return null if the string is null
     let trim (value: string) = match value with null -> null | x -> x.Trim()
```
vasily-kirichenko: You can use `==` here:
```fsharp
let trim (value: string) = if value == null then null else x.Trim()
```

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -517,28 +561,33 @@ module String =
...
     
+    /// Splits a string into substrings based on the strings in the array seperators
     let split options (separator: string[]) (value: string) = 
         match value with null -> null | x -> x.Split(separator, options)
```
vasily-kirichenko: Same as above.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -37,51 +39,50 @@ type QuickInfoMargin (textDocument: ITextDocument,
...
+            let errorString (errors:string list) (sb:StringBuilder) =
+                match errors with 
+                | [e] -> append e sb
+                | _ -> (sb, errors ) ||> List.foldi (fun sb i e -> 
+                    sb |> append (sprintf "%d. %s" (i + 1) e) |> append " ")
+                       
+            let currentInfo =
+                // if the tooltip contains errors show them
+                errors |> Option.map (fun errors ->
+                    (StringBuilder (), errors) ||> Array.fold (fun sb (severity, err) ->
```
vasily-kirichenko: Why folding if we can use plane loop here? The `sb` is mutable anyway, no profit from fold, but less readable.

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
dungpa: We have `Option.orTry` function to do that. We haven't used this as much as we should.

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine: whoops should be 0.8x slower

dungpa:
Since this PR introduces many functions implemented in an imperative way, please add FsCheck tests for those. See what Vasily did in this PR https://github.com/fsprojects/VisualFSharpPowerTools/pull/1186

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -3,6 +3,24 @@
...
+    /// obj.ReferenceEquals
+    let inline (==) a b = obj.ReferenceEquals(a, b)
+    /// LanguagePrimitives.PhysicalEquality
+    let inline (===) a b = LanguagePrimitives.PhysicalEquality a b
+    let inline debug msg = Printf.kprintf Debug.WriteLine msg
+    let inline fail msg = Printf.kprintf Debug.Fail msg
+    let inline isNull v = match v with | null -> true | _ -> false
+    let inline isNotNull v = v |> (not << isNull)
+    let inline dispose (disposable:#IDisposable) = disposable.Dispose ()
+    let inline (+=) (a1:'a byref) (a2:'a) = a1 <- a1 + a2
```
dungpa: I don't see the point of `+=` and `-=`. I prefer explicit mutation to some uncommon assignment constructs that I have to look up.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -3,6 +3,24 @@
...
+    let inline (-=) (a1:'a byref) (a2:'a) = a1 <- a1 - a2
+
+    type 'a List with
+        member x.Last = x.[x.Length-1]
```
dungpa: Please don't do this. Why traversing the whole list twice to get the last element? You can do it by a recursive function and pattern matching.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -27,60 +48,136 @@ module List =
...
+
     /// Returns true if one array has another as its subset from index 0.
-    let startsWith (prefix: _ []) (whole: _ []) =
-        let rec loop index =
-            if index > prefix.Length - 1 then true
-            elif index > whole.Length - 1 then false
-            elif prefix.[index] = whole.[index] then loop (index + 1)
-            else false
-        prefix.Length = 0 || loop 0
+    let startsWith (prefix: _ []) (whole: _ []) =
```
dungpa: Why this function is better than previous implementation? It allocates a new array in each iteration which was the point to introduce the previous implementation?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -27,60 +48,136 @@ module List =
...
-            else false
-
-        if suffix.Length = 0 then true
-        elif whole.Length < suffix.Length then false
-        else loop 1
+        if isNull suffix || isNull whole then false
+        elif suffix.Length = 0 then true
+        elif suffix.Length > whole.Length then false
+        elif suffix.Length = whole.Length then areEqual suffix whole
+        else areEqual whole.[whole.Length-suffix.Length..whole.Length-1] suffix
```
dungpa: Good refactoring :+1: .

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -581,6 +739,45 @@ module String =
...
+        loop (reader.ReadLine())
+
+open System.Text
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module StringBuilder =
+    /// Pipelining function for appending a string to a stringbuilder
+    let inline append   (str:string) (sb:StringBuilder) = sb.Append str
+
+    /// Pipelining function for appending a string with a '\n' to a stringbuilder
+    let inline appendln (str:string) (sb:StringBuilder) = sb.AppendLine str
```
dungpa: For consistency, it's better to be `appendLine`.

src/FSharpVSPowerTools.Core/Utils.fs
vasily-kirichenko: :+1:

src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
```diff
@@ -174,38 +173,45 @@ type SyntaxConstructClassifier
...
+                    yield newSpans
+                }   
+                |> Array.concat 
+                |> Array.sortBy (fun x -> x.WordSpan.Line)
+
+        match oldSpans , newSpans with
+        | null, null | [||], [||] -> [||]
+        | [||], nspans   -> nspans 
+        | ospans , [||]  -> ospans
+        | ospans, nspans -> compareSpans ospans nspans
```
vasily-kirichenko: What about `null, [||]` and `[||], null`?

src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
```diff
@@ -174,38 +173,45 @@ type SyntaxConstructClassifier
...
+            else
+                debug "[SyntaxConstructClassifier] Mergin spans (new range %A < old range %A)."
+                      (newStartLine, newEndLine) (oldStartLine, oldEndLine)
+                seq {
+                    yield (oldSpans |> Array.takeWhile (fun x -> x.WordSpan.Line < newStartLine))
+                    yield (oldSpans |> Array.skipWhile (fun x -> x.WordSpan.Line < newEndLine)) 
+                    yield newSpans
+                }   
+                |> Array.concat 
+                |> Array.sortBy (fun x -> x.WordSpan.Line)
```
vasily-kirichenko: I don't think arrays are more preferable here. The old implementation does not allocate additional arrays.

src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
```diff
@@ -447,27 +451,24 @@ type SyntaxConstructClassifier
...
+                    let! span = fromRange origSnapshot (columnSpan.WordSpan.ToRange())
+                    let span = 
+                        if targetSnapshotSpan.Snapshot <> span.Snapshot then
+                            span.TranslateTo(targetSnapshotSpan.Snapshot, SpanTrackingMode.EdgeExclusive)  
+                        else span
+                    // Translate the span to the new snapshot
+                    return clType, span 
+                })
+            |> Array.takeWhile (fun (_, span) -> span.Start.GetContainingLine().LineNumber <= spanEndLine)
+            |> Array.map (fun (clType, span) -> ClassificationSpan (span, clType))
```
vasily-kirichenko: Oh, man... This whole pipeline must be lazy (== `Seq`). Otherwise, we will `translate` all spans to new snapshot, which is very expensive. Please, revert it.

src/FSharpVSPowerTools.Core/Utils.fs
vasily-kirichenko: :+1:

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -27,60 +48,136 @@ module List =
...
+                    prev.Add v
+                else 
+                    let prev = new ResizeArray<'T> 1
+                    dict.[key] <- prev
+                    prev.Add v
+                loop t
+            | _ -> ()
+        loop list
+        dict  // Return the list-of-lists.
+        |> Seq.map (fun group -> (group.Key, group.Value.ToArray() |> List.ofArray))
```
vasily-kirichenko: Why `group.Value.ToArray() |> List.ofArray` is necessary? What about `List.ofSeq group.Value`?

src/FSharpVSPowerTools.Core/Utils.fs
vasily-kirichenko: :+1:

src/FSharpVSPowerTools.Core/Utils.fs
vasily-kirichenko: It allocates a new array in the last line for no reason. How it's better?..

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -89,32 +186,94 @@ module Array =
...
+    let distinctBy keyf (array:'T[]) =
+        checkNonNull "array" array
+        let temp = Array.zeroCreate array.Length
+        let mutable i = 0 
+        let hashSet = HashSet<_> HashIdentity.Structural<_>
+        for v in array do
+            if hashSet.Add(keyf v) then
+                temp.[i] <- v
+                &i += 1
+        temp.[0..i]
```
vasily-kirichenko: Maybe it's better to use `ResizeArray` here and get rid of fragile logic around `i`?

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine: I had a feeling you guys wouldn't be fans, but I thought it was worth a shot.

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine: let inline stringArr x = Array.init x string

let last (x:'a list) = x.[x.Length-1]

let rec rlast (x:'a list) = 
    match x with 
    | [] -> failwith "empty has no last"   
    |l::[] -> l
    |hd::tl -> rlast tl

;;

let test data = compareTwoRuntimes 100 "last" (fun _ -> last data|>ignore) "recursive last" (fun _ -> rlast data|> ignore);;

let d0 = stringArr 1000     |> Array.toList
let d1 = stringArr 10000    |> Array.toList
let d2 = stringArr 100000   |> Array.toList
let d3 = stringArr 1000000  |> Array.toList
;;

test d0 ;;
test d1 ;;
test d2 ;;
test d3 ;;

vasily-kirichenko: The function is List.last, not Array.last :)

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine:
```
> original startswith 0.02ms
startswithIf 0.02ms
Ran 100 times
Ratio:  1.0
val it : unit = ()
>
original startswith 0.15ms
startswithIf 0.1ms
Ran 100 times
Ratio:  1.5
val it : unit = ()
> original startswith 20.36ms
startswithIf 11.89ms
Ran 100 times
Ratio:  1.712363331
val it : unit = ()
>
startswithMatch 0.06ms
startswithIf 0.04ms
Ran 100 times
Ratio:  1.5
val it : unit = ()
> startswithMatch 0.15ms
startswithIf 0.07ms
Ran 100 times
Ratio:  2.142857143
val it : unit = ()
> startswithMatch 11.92ms
startswithIf 12.38ms
Ran 100 times
Ratio:  0.9628432956
val it : unit = ()
```
test code compiled in release mode https://gist.github.com/cloudRoutine/88143149f02f766b1389

src/FSharpVSPowerTools.Core/Utils.fs
dungpa: Here are results on my machines:

> last 0.004ms
recursive last 0.002ms
  Ratio:  2.0
val it : unit = ()
> last 0.037ms
recursive last 0.021ms
  Ratio:  1.761904762
val it : unit = ()
> last 0.383ms
recursive last 0.222ms
  Ratio:  1.725225225
val it : unit = ()
> last 4.576ms
recursive last 2.495ms
  Ratio:  1.834068136
val it : unit = ()
>

It sounds reasonable wrt the dominant cost of list traversal (2:1 between the two implementations).

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine: I used the FSharp.Core implementation
https://github.com/Microsoft/visualfsharp/blob/fsharp4/src/fsharp/FSharp.Core/array.fs#L278-L287

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine: I ran the tests a few more times and I got results like yous, I don't know what that first set was about :confused:

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine:
```
groupby seq 0.55ms
groupby arr 0.14ms
Ran 100 times
Ratio:  3.928571429
groupby seq 1.55ms
groupby arr 1.7ms
Ran 100 times
Ratio:  0.9117647059
groupby seq 20.19ms
groupby arr 21.12ms
Ran 100 times
Ratio:  0.9559659091
groupby seq 199.47ms
groupby arr 212.42ms
Ran 100 times
Ratio:  0.9390358723
```
It seemed like we'd be using it within the bounds where using an array internally was faster
[full code](https://gist.github.com/cloudRoutine/ada3b853497749024411)

cloudRoutine: that was a bad key function
with `let inline keyfn (str:string) = str.Length % 4`
Seq is better 
```
groupby seq 0.13ms
groupby arr 0.41ms
Ran 100 times
Ratio:  0.3170731707
groupby seq 1.77ms
groupby arr 1.89ms
Ran 100 times
Ratio:  0.9365079365
groupby seq 21.36ms
groupby arr 18.18ms
Ran 100 times
Ratio:  1.174917492
groupby seq 203.43ms
groupby arr 202.44ms
Ran 100 times
Ratio:  1.004890338
```

src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
cloudRoutine: switching this back made it slower, but smooth again. smooth definitely more important

vasily-kirichenko: How have you determined that it "made it slower"?

cloudRoutine: the amount of time it takes to render the syntax coloring and populate the outlining tags

src/FSharpVSPowerTools.Core/Utils.fs
dungpa: I've got similar results on my machine. Notice that the new implementation incurs more GC than the old one.

src/FSharpVSPowerTools.Core/InterfaceStubGenerator.fs
```diff
@@ -401,7 +401,7 @@ module InterfaceStubGenerator =
 
     let internal (|LongIdentPattern|_|) = function
         | SynPat.LongIdent(LongIdentWithDots(xs, _), _, _, _, _, _) ->
-            let (name, range) = xs |> Seq.map (fun x -> x.idText, x.idRange) |> Seq.last
+            let (name, range) = xs |> List.map (fun x -> x.idText, x.idRange) |> List.last
```
vladima: as I see you ultimately want to get the last element of the list and then take its constituents. So why do you need to map the entire sequence if you are ignoring results of projection. To me it should be smth like
```fsharp
    let last = List.last xs
    Some (last.idText, last.idRange)
```

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -89,32 +186,94 @@ module Array =
...
+        let hashSet = HashSet<_> HashIdentity.Structural<_>
+        for v in array do
+            if hashSet.Add(keyf v) then
+                temp.[i] <- v
                 i <- i + 1
-            result
+        temp.[0..i]
+
+    /// pass an array byref to reverse it in place
+    let revInPlace (arr: 'a[] byref ) =
```
vladima: given that array is already reference type and this function does not change the identify of the parameter what is the point of making it byref?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -545,14 +696,18 @@ module Dict =
...
     let lowerCaseFirstChar (str: string) =
-        match str with
-        | null -> null
-        | str ->
-            match str.ToCharArray() |> Array.toList with
-            | [] -> str
-            | h :: t when Char.IsUpper h -> String (Char.ToLower h :: t |> List.toArray)
-            | _ -> str
+        if isNull str then null else 
+        let strArr = toCharArray str
```
vladima: what about checking if string can be returned as is before doing array allocation?
```fsharp
if String.IsNullOrEmpty str || Char.IsLower(str, 0) then str
else 
// do the rest 
```

src/FSharpVSPowerTools.Logic/QuickInfoMargin.fs
```diff
@@ -34,57 +36,92 @@ type QuickInfoMargin (textDocument: ITextDocument,
        visual.tbQuickInfo.MouseDoubleClick.Add (fun _ ->
            System.Windows.Clipboard.SetText visual.tbQuickInfo.Text
            visual.tbQuickInfo.SelectAll())
-    
+
     let buffer = view.TextBuffer
     let mutable currentWord: SnapshotSpan option = None
 
+<<<<<<< HEAD
```
vladima: merge marker :)

src/FSharpVSPowerTools.Core/InterfaceStubGenerator.fs
vasily-kirichenko: :+1:

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -545,14 +700,18 @@ module Dict =
...
+
     let lowerCaseFirstChar (str: string) =
-        match str with
-        | null -> null
-        | str ->
-            match str.ToCharArray() |> Array.toList with
-            | [] -> str
-            | h :: t when Char.IsUpper h -> String (Char.ToLower h :: t |> List.toArray)
-            | _ -> str
+        if isNull str then null else
```
vasily-kirichenko: This pattern (`if isNull ... then null else ...`) repeats quite often now. What about:

```fsharp
[<RequireQualifiedAccess>]
module Null =
    let inline fill defaultValue x = if isNull x then null else defaultValue
    let inline fillWith (genDefaultValue: unit -> 'a) x = if isNull x then null else genDefaultValue()
```
Then you can rewrite the code like this:
```fsharp
str |> Null.fill (fun _ -> ...)
```

cloudRoutine: What's the value of all of this `null` preservation that we're doing? If we want strings and we get null we could treat it as an empty string, if we get `null` when we wanted an array then `[||]`, etc.

cloudRoutine:
I think I've finally addressed all the issues

src/FSharpVSPowerTools.Core/SignatureGenerator.fs
```diff
@@ -455,11 +455,11 @@ and internal writeModuleHeader ctx (modul: FSharpEntity) =
     let qualifiedModuleName = tryRemoveModuleSuffix modul modul.FullName
     writer.WriteLine("module {0}", qualifiedModuleName)
     ctx.ResolvingOpenDeclarations
-    |> Seq.choose (fun (openDecl, isUsed) -> if isUsed then Some openDecl else None)
-    |> Seq.iteri (fun i decl ->
+    |> Array.iteri (fun i (openDecl, isUsed) -> 
+        if not isUsed then () else
```
vasily-kirichenko: Remove negative branch.

src/FSharpVSPowerTools.Core/SignatureGenerator.fs
```diff
@@ -471,13 +471,13 @@ and internal getParentPath (entity: FSharpEntity) =
 and internal writeTypeHeader ctx (typ: FSharpEntity) =
     let writer = ctx.OpenDeclWriter
     let parent = getParentPath typ
-    writer.WriteLine(parent)
+    writer.WriteLine parent
     ctx.ResolvingOpenDeclarations
-    |> Seq.choose (fun (openDecl, isUsed) -> if isUsed then Some openDecl else None)
-    |> Seq.iteri (fun i decl ->
+    |> Array.iteri (fun i (openDecl, isUsed) -> 
+        if not isUsed then () else
```
vasily-kirichenko: Same as above.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -27,94 +58,241 @@ module List =
...
+        | _ -> 
+            let f = OptimizedClosures.FSharpFunc<_,_,_,_>.Adapt folder
+            let rec loop idx s xs = 
+                match xs with 
+                | [] -> s
+                | h::t -> loop (idx+1) (f.Invoke(s,idx,h)) t
+            loop 0 state xs
+
+
+    let inline groupBy (keyfn:'T->'Key) (list: 'T list) =
```
vasily-kirichenko: Why it's inlined but `foldi`, for example, is not?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -27,94 +58,241 @@ module List =
...
+        let hashSet = HashSet<_> HashIdentity.Structural<_>
+        for v in array do
+            if hashSet.Add(keyf v) then
+                temp.[i] <- v
                 i <- i + 1
-            result
+        temp.[0..i-1]
+
+    /// pass an array byref to reverse it in place
+    let revInPlace (array: 'a[] byref ) =
```
vasily-kirichenko: You don't have to pass array by ref because arrays are ref types. Change it to `let revInPlace (array: 'a[])`.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -545,46 +714,52 @@ module Dict =
...
 
-    let trim (value: string) = match value with null -> null | x -> x.Trim()
+    /// Remove all trailing and leading whitespace from the string
+    /// return null if the string is null
+    let trim (value: string) = if isNull value then null else value.Trim()
     
+    /// Splits a string into substrings based on the strings in the array seperators
     let split options (separator: string[]) (value: string) = 
-        match value with null -> null | x -> x.Split(separator, options)
+        if isNull value  then null else value.Split(separator, options)
```
vasily-kirichenko: What about `Null.map` or something like this?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -613,6 +788,45 @@ module String =
...
 
+    open System.Text
+    /// Use an accumulation function to create a new string applying a transformation
+    /// to every non-empty line in the string
+    let mapNonEmptyLines (folder: StringBuilder -> string -> StringBuilder) (str: string) =
+        let f = OptimizedClosures.FSharpFunc<_,_,_>.Adapt folder
+        use reader = new StringReader (str)
+        let sb = StringBuilder ()
+        let mutable line = reader.ReadLine ()
+        while isNotNull line do
```
vasily-kirichenko: Why not use a recursive function and get rid of the mutable var?

src/FSharpVSPowerTools.Core/Utils.fs
cloudRoutine: you want to call `Null.fill` , `Null.map` instead? otherwise I'm not sure what you mean.
```
    let split2 options (separator: string[]) (value: string) = 
        Null.fill value (value.Split(separator, options))
```

vasily-kirichenko:
```null => null
x => f x```
is not `fill`, it's `map`.

Comment (cloudRoutine) [inline src/FSharpVSPowerTools.Core/Utils.fs]: I think this function might invite mistakes, if someone swaps the order by mistake, and since we keep passing nulls through everywhere it'd be hard to identify the origin

I think we should scrap those `Null` functions and leave it explicit for the sake of sanity

Comment (cloudRoutine) [inline src/FSharpVSPowerTools.Core/Utils.fs]:
```
module Null =
    let map fn value =
        if isNull value then null else fn value

let trim (value: string) = value |> Null.map (fun x-> x.Trim())
```

vasily-kirichenko: Why we do not convert all `null`s into `Option` on VS boundaries?

cloudRoutine: ^ this I like, and `Null.fillWith` is fine, but I still think `Null.fill` is iffy at best

cloudRoutine: I've been converting nulls to options and the VS boundaries for the services  as I convert it all into F#. I don't think we need to tackle that for the entire codebase on this PR

PR #1224, opened 2015-11-01
Add support to change the tooltip zoom level for outlining

zhengbli:
capture

### DIFF
--- a/src/FSharpVSPowerTools.Logic/Outlining/OutliningTagger.fs
+++ b/src/FSharpVSPowerTools.Logic/Outlining/OutliningTagger.fs
@@ -176,7 +176,7 @@ type OutliningTagger
     let createElisionBufferView (textEditorFactoryService: ITextEditorFactoryService) (finalBuffer: ITextBuffer) =
         let roles = textEditorFactoryService.CreateTextViewRoleSet ""
         let view = textEditorFactoryService.CreateTextView (finalBuffer, roles, Background = Brushes.Transparent)
-        let zoomLevel = outliningOptions.Value.TooltipZoomLevel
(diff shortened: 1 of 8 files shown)


vasily-kirichenko:
If I enter `0.8` it does not complain, just reset in silently to `0` which is not good. It should highlight the text box in red and / or shows modal dialog if user tries to close settings dialog having invalid zoom level (I believe we do just this for "Navigation bar" setting if VS is running without admin privileges). 

I think it's better to set zoom level in percents, like the editor shows its zoom level:

image

vasily-kirichenko:
On my machine `0,9` is OK, `0.9` is not (it's reset to `0`). So if you are gonna keep supporting decimals, you should parse both formats.

vasily-kirichenko:
Also it should be impossible to enter / save zoom level outside `1%..100%`.

zhengbli:
I changed it accordingly, not it looks more like VS's zoom setting:

vasily-kirichenko:
image

zhengbli:
Hmmm, could it be caching issues? I just uninstalled and reinstalled on my machine, it runs fine.

zhengbli:
I'll try on a new machine on Monday.

vasily-kirichenko:
I deleted `c:\Users\kot\AppData\Local\Microsoft\VisualStudio\12.0Exp` and `c:\Users\kot\AppData\Local\Microsoft\VisualStudio\Exp` - same result.

zhengbli:
oh shoot, it might be that I used nullable int type that breaked vs2013 (I only have 15 installed). Will fix soon

PR #1225, opened 2015-11-01
fixed: SyntaxConstructClassifier uses old spans after deleting lines somewhere in the middle of file

vasily-kirichenko:
It's becoming really tricky :(


src/FSharpVSPowerTools.Logic/SyntaxConstructClassifier.fs
```diff
@@ -175,29 +175,53 @@ type SyntaxConstructClassifier
...
-                    | Category.Quotation 
-                    | Category.Unused -> false
-                    | _ -> true) 
-                |> Array.map (fun x -> x.WordSpan.Line)
-            Array.min lines, Array.max lines
+            let typeCheckerSpans = getTypeCheckerSpans spans
+            typeCheckerSpans,
+            typeCheckerSpans
+            |> Array.map (fun x -> x.WordSpan.Line)
+            |> function [||] -> -1, -1 | lines -> Array.min lines, Array.max lines
```
dungpa: This `-1` thing is convenient but dangerous. There is no tracking from the type system so it's easy to misuse it.

PR #1233, opened 2015-11-05
SyntaxConstructClassifier caches line number for each span

vasily-kirichenko:
WIP

Scenario: select large block (`Shift+Alt+arrows`), which is not fit into editor visible area (~50-100 lines or larger). At some point cursor almost stops moving and CPU is high loaded.


dungpa:
Great.

PR #1234, opened 2015-11-05
Remove integration tests

dungpa:
We wanted to write integration tests and ran them directly inside VS. However, this infrastructure is not useful; we haven't even been able to write smoke tests for the extension. Removing for good.


vasily-kirichenko:
The less code the better :+1:

PR #1237, opened 2015-11-07
Optimize outlining

vasily-kirichenko:
It filters out outlining spans of disabled (in settings) scopes on earlier stage to improve performance. The main change is over here https://github.com/fsprojects/VisualFSharpPowerTools/compare/fsprojects:master...vasily-kirichenko:optimize-outlining?expand=1#diff-d769655d7e66f49c8169b86d6aa975f1R174

### DIFF
--- a/src/FSharpVSPowerTools.Logic/EventHandling.fs
+++ b/src/FSharpVSPowerTools.Logic/EventHandling.fs
@@ -3,7 +3,6 @@
 open System
 open Microsoft.VisualStudio.Text
 open Microsoft.VisualStudio.Text.Editor
-open Microsoft.FSharp.Compiler.Range
 open System.Threading
 open System.Windows.Threading
 open FSharpVSPowerTools
@@ -65,9 +64,9 @@ type DocumentEventListener (events: IEvent<unit> list, delayMillis: uint16, upda
                     startNewTimer()
                     do! awaitPauseAfterChange()
                 protectUpdate() }
-       Async.StartImmediateSafe(computation, tokenSource.Token)
        // Go ahead and synchronously get the first bit of info for the original rendering
        protectUpdate()
+       Async.StartInThreadPoolSafe (computation, tokenSource.Token)
 
     /// Skip all timer events in order to test events instantaneously
     static member internal SkipTimerDelay 
--- a/src/FSharpVSPowerTools.Logic/EventHandling.fs
+++ b/src/FSharpVSPowerTools.Logic/EventHandling.fs
@@ -14,7 +14,7 @@ module ViewChange =
     let layoutEvent (view: ITextView) = 
         view.LayoutChanged |> Event.choose (fun e -> if e.NewSnapshot <> e.OldSnapshot then Some() else None)
     
-    let viewportHeightEvent (view: ITextView) = 
+    let viewportHeightEvent (view: ITextView) =  
         view.ViewportHeightChanged |> Event.map (fun _ -> ())
 
     let caretEvent (view: ITextView) = 
@@ -64,8 +64,12 @@ type DocumentEventListener (events: IEvent<unit> list, delayMillis: uint16, upda
                     startNewTimer()
                     do! awaitPauseAfterChange()
                 protectUpdate() }
-       // Go ahead and synchronously get the first bit of info for the original rendering
-       protectUpdate()
+       // Go ahead and get the first bit of info for the original rendering
+       let uiContext = SynchronizationContext.Current
+       Async.StartInThreadPoolSafe <| async {
+           do! Async.SwitchToContext uiContext
+           protectUpdate()
+       }
        Async.StartInThreadPoolSafe (computation, tokenSource.Token)
 
     /// Skip all timer events in order to test events instantaneously
--- a/src/FSharpVSPowerTools.Logic/EventHandling.fs
+++ b/src/FSharpVSPowerTools.Logic/EventHandling.fs
@@ -26,7 +26,19 @@ module ViewChange =
     let tagsEvent (tagAggregator: ITagAggregator<_>) = 
         tagAggregator.TagsChanged |> Event.map (fun _ -> ())
 
-type DocumentEventListener (events: IEvent<unit> list, delayMillis: uint16, update: unit -> unit) =
+[<NoComparison; NoEquality>]
+type CallInUIContext = CallInUIContext of ((unit -> unit) -> Async<unit>)
+    with static member FromCurrentThread() = 
+                         let uiContext = SynchronizationContext.Current
+                         CallInUIContext (fun f ->
+                             async {
+                                 let ctx = SynchronizationContext.Current
+                                 do! Async.SwitchToContext uiContext
+                                 protect f
+                                 do! Async.SwitchToContext ctx
+                             })
+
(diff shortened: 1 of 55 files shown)


vasily-kirichenko:
Profiling code editing on `TypeChecker.fs` showed that `ITextSnapshot.GetText()` took 98% time of UI thread. So I added caching each file text in `OpenDocumentTracker`. 

I have not checked carefully that I removed all `GetText()` occurrences yet.

dungpa:
Why do you switch the first run to be async on UI thread? It has to be run on UI thread anyway, so why bother?

vasily-kirichenko:
If it runs on UI thread immediately, `OpenDocumentTracker` does not receive `RegisterView` call yet and its `TryGetDocumentText` returns `None`. This code

```fsharp
let uiContext = SynchronizationContext.Current
       Async.StartInThreadPoolSafe <| async {
           do! Async.SwitchToContext uiContext
           protectUpdate()
       }
```
allows UI thread to finish constructing all providers and notifying `OpenDocumentTracker` about the newly opened Buffer.

vasily-kirichenko:
@dungpa I've fixed the tests. Please review.

src/FSharpVSPowerTools.Logic/EventHandling.fs
```diff
@@ -65,9 +64,13 @@ type DocumentEventListener (events: IEvent<unit> list, delayMillis: uint16, upda
                     startNewTimer()
                     do! awaitPauseAfterChange()
                 protectUpdate() }
-       Async.StartImmediateSafe(computation, tokenSource.Token)
-       // Go ahead and synchronously get the first bit of info for the original rendering
-       protectUpdate()
+       // Go ahead and get the first bit of info for the original rendering
```
dungpa: We don't think we get this part right. The callback function is now run in thread pools. Remember that we had to revert similar changes in this https://github.com/fsprojects/VisualFSharpPowerTools/pull/1141in order to avoid deadlocks.

If you would like to make it async on UI thread, we could do:

```fsharp
    do 
       let computation =
           async { 
            // Go ahead and get the first bit of info for the original rendering
            protectUpdate()
            while true do
                do! Async.AwaitEvent events
                if not skipTimerDelay then
                    startNewTimer()
                    do! awaitPauseAfterChange()
                protectUpdate() }
       Async.StartImmediateSafe (computation, tokenSource.Token)
```

vasily-kirichenko:
Your code is the same as the old one - it calls first `protectUpdate` immediately, during `DocumentEventListener` construction (because `Async.StartImmediate (async { doSomething() })` is the same as `doSomething` since `StartImmediate` performs everything in `async` block, up to first `let!`, `do!` or `return!`, in the thread it was called). I.e. the following sequence is executed in UI thread without any breaks, sequentially and blocking:

* User open a document
* VS instantiates all our taggers, filters etc.
* The taggers create (in their constructors) instances of `DocumentEventListener`
* `DocumentEventListener`s (in their constructors) call `protectUpdate`
* `protectUpdate`s do heavy things like getting `IProjectProvider`, `ITextBuffer.GetText()` and even `VSLanguageService.GetSymbol()`

So until *all* these stuff finish its work, UI is frozen. 

New code 

```fsharp
let computation =
    async { 
     while true do
         do! Async.AwaitEvent events
         if not skipTimerDelay then
             startNewTimer()
             do! awaitPauseAfterChange()
         protectUpdate() }
// Go ahead and get the first bit of info for the original rendering
let uiContext = SynchronizationContext.Current
Async.StartInThreadPoolSafe <| async {
    do! Async.SwitchToContext uiContext
    protectUpdate()
}
Async.StartInThreadPoolSafe (computation, tokenSource.Token)
```

works differently:

* VS creates taggers, they create `DocumentEventListener`s
* `DocumentEventListener`s starts `async`s in `ThreadPool`, which *will* switch to UI thread and perform the heavy work
* `DocumentEventListener`s starts the old good infinite loop waiting VS events in `ThreadPool`
* UI thread is freed at this point which means that documents are open faster

So the new behavior has very clear benefits and works differently than the dead-locking one from that reverted PR. 

Actually code editing experience is very nice now, even on `TypeChecker.fs`. 
I'll use this version tomorrow at work at will see how it works in real world.

dungpa:
I see the point of scheduling the first `protectUpdate` on thread pools then switch to UI thread and execute on UI thread. 

But why don't we do the same for `protectUpdate` in the `while` loop? In other words, why aren't `do! Async.SwitchToContext uiContext` in the loop? What is the reason for this inconsistency?

vasily-kirichenko:
Well, we don't know what VS API we must call from UI thread only. For some reason we assume that the following code:

```fsharp
maybe {
    let dte = serviceProvider.GetService<EnvDTE.DTE, SDTE>()
    let! doc = dte.GetCurrentDocument(textDocument.FilePath)
    let! project = projectFactory.CreateForDocument buffer doc 
    return currentRequest, doc, project 
}
```
, which is copy-pasted in every tagger, must be called from UI thread. Yes, when I called it from ThreadPool threads VS behavied like UI thread was waiting something forever (not working UI, 0% CPU). We should clean up our understanding out this stuff. I remember Valdimir Matveev said something about it.

vasily-kirichenko:
I removed switching to UI thread for the first `protectUpdate` call and got the following exception:

```
[VFPT][Error 05:24:41 ] Exception Message: The calling thread cannot access this object because a different thread owns it.
Stack Trace:    at System.Windows.Threading.Dispatcher.VerifyAccess()
   at System.Windows.DependencyObject.GetValue(DependencyProperty dp)
   at System.Windows.Controls.Panel.get_IsItemsHost()
   at System.Windows.Controls.UIElementCollection.VerifyWriteAccess()
   at System.Windows.Controls.UIElementCollection.Clear()
   at <StartupCode$FSharpVSPowerTools-Logic>.$UnusedDeclarationMargin.clo@28-159.Invoke(Unit _arg2)
   at FSharpVSPowerTools.ProjectSystem.VSUtils.protectOrDefault[a](FSharpFunc`2 f, a defaultVal)
```

dungpa:
We could add `ForegroundThreadGuard.CheckThread` https://github.com/fsprojects/VisualFSharpPowerTools/blob/85f4aa175025ec71fd4b44e79ba2b52b910a9176/src/FSharpVSPowerTools.Logic/VSUtils.fs#L383 to the start of every `updateAtCaretPosition` function (e.g. https://github.com/fsprojects/VisualFSharpPowerTools/blob/85f4aa175025ec71fd4b44e79ba2b52b910a9176/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs#L81) to avoid accidental refactoring.

Note that for some features, it's fine with calling from thread pool. Here are the best references I could find https://msdn.microsoft.com/en-us/library/dn448839.aspx or https://github.com/Microsoft/VSProjectSystem/blob/94bd2eac351477d37a4e2a80d32e73a4e1bbd347/doc/overview/3_threading_rules.md. cc @vladima in case he has a clear explanation for this.

vasily-kirichenko:
Please review the last commit. `DocumentEventListener`:

* cancel previous invocation of callback
* starts callback in freshly created async
* passes `CallInUIThread` function into the callback given it an opportunity to call in UI thread whatever it needs.

dungpa:
Great solution. :+1:

PR #1238, opened 2015-11-08
Treat warnings as errors

dungpa:
For some reason, my previous PR only applies warnings as errors for debug configuration. This PR reinforces it for other configurations + add a few minor tweaks.


vasily-kirichenko:
Great. Yes, Setting "treat warnings as errors" and "warn level" is tricky.

PR #1245, opened 2015-11-12
clean up NavigateToItem and others

vasily-kirichenko:
yield results from project check tasks as they finish, not one-by-one
add a untyped ast cache (wip)

Performance difference is *huge*.


vasily-kirichenko:
Works. I cache `NavigableItems` per file, excluding open files. It's simple, but seems to work very nicely.

dungpa:
> Performance difference is huge.

How huge?

I think core tests need to be updated in order that the solution compiles.

You can try using `[<ImportingConstructor>]` instead of `[<Import; DefaultValue>]`. The code might be cleaner.

vasily-kirichenko:
Before (three runs just after solution is loaded and a file fully colorized):



After:

vasily-kirichenko:
It would be even cooler if we

* warm up the cache just after solution is loaded to make the very first run as fast as subsequent ones
* store cache content into a file (`fileName * fileWriteTime * NavigableItem[]) in order to prevent reparsing everything on first run

vasily-kirichenko:
Saving/loading performance on FCS solution is as following:

Loaded: 26403 items for 179 files. Sutable: 26353 items for 178 files. Elapsed 00:00:00.2595570

Saved 26403 items for 179 files in 00:00:00.1151705

vasily-kirichenko:
It's ready.

src/FSharpVSPowerTools.Logic/NavigableItemCache.fs
```diff
@@ -0,0 +1,139 @@
...
+
+    let cacheFilePath solutionPath =
+        let solutionHash = solutionPathHash solutionPath
+        let root = Environment.ExpandEnvironmentVariables @"%LOCALAPPDATA%\VisualFSharpPowerTools\SolutionCaches"
+        root </> solutionHash </> "navigable_items.cache"
+
+    let loadFromDisk (solutionPath: FilePath) =
+        VSUtils.protect <| fun _ ->
+            cache.Clear()
+            let sw = Stopwatch.StartNew()
```
dungpa: Please use `Profiler` type from Core instead of `Stopwatch`. It will make it easier if we ever remove profiling code.

src/FSharpVSPowerTools.Logic/NavigableItemCache.fs
```diff
@@ -0,0 +1,139 @@
...
+            Logging.logInfo <| fun _ -> 
+                sprintf "[NavigableItemCache] Loaded: %d items for %d files. Sutable: %d items for %d files. Elapsed %O" 
+                        (items |> Array.sumBy (fun x -> x.Items.Length)) items.Length
+                        (cache.Values |> Seq.sumBy (fun x -> x.Items.Length)) cache.Count
+                        sw.Elapsed
+
+    let saveToDisk (solutionPath: FilePath) =
+        if !dirty then
+            dirty := false
+            VSUtils.protect <| fun _ ->
```
dungpa: `VSUtils` is auto-open, so it's not needed here.

src/FSharpVSPowerTools.Logic/NavigableItemCache.fs
```diff
@@ -0,0 +1,139 @@
...
+                let sw = Stopwatch.StartNew()
```
dungpa: Same comment. Please use `Profiler` instead.

src/FSharpVSPowerTools.Logic/NavigateToItem.fs
```diff
@@ -81,166 +156,153 @@ and
...
+type NavigateToItemProviderFactory 
+    [<ImportingConstructor>]
+    (
+        openDocumentsTracker: IOpenDocumentsTracker,
+        [<Import(typeof<SVsServiceProvider>)>] serviceProvider: IServiceProvider,
+        languageService: VSLanguageService,
+        [<ImportMany>] itemDisplayFactories: seq<Lazy<INavigateToItemDisplayFactory, IMinimalVisualStudioVersionMetadata>>,
+        vsCompositionService: ICompositionService,
+        projectFactory: ProjectFactory,
+        [<Import(typeof<NavigableItemCache>)>] navigableItemCache: NavigableItemCache
```
dungpa: An unintended effect of importing this cache may be if even we disable the feature, the cache is still initialized and hooked with VS events. I think it's better to make it a lazy value and call its constructor if the feature is enabled. 

Note that we will only have a single instance of `NavigateToItemProviderFactory` by MEF.

vasily-kirichenko:
Thanks, all done.

dungpa:
Very cool.

PR #1248, opened 2015-11-14
Protect against exceptions in FsiReferenceCommand

dungpa:
Saw the conversation on Slack. I thought this should be done just in case.


vasily-kirichenko:
Great.

PR #1250, opened 2015-11-15
SyntaxConstructClassifier accepts new spans if there are no parse or check errors even though the new range is smaller than old one

vasily-kirichenko:


dungpa:
:+1:

PR #1252, opened 2015-11-16
fixed: AssemblyContentProvider can return None if some of entities th…

vasily-kirichenko:
…rows on MembersFunctionsAndValues property getter

streamline and fix ResolveUnopenedNamespaceTagger logic

I'm surprised it's been working (sort of) all this time :(


vasily-kirichenko:
The exception `MembersFunctionsAndValues` raised was like this:

image

dungpa:
I thought we know when FCS throws exceptions, but apparently we don't.

PR #1254, opened 2015-11-18
Language service tests refactoring

vasily-kirichenko:
Rewrite `LanguageServiceTests` in the same way `SourceCodeClassifierTests` is written (eliminate  using of `LanguageServiceSampleFile.fs`, use inlined small snippets instead).


dungpa:
Nice. It would be great to eventually get rid of LanguageServiceSampleFile.fs :-)

vasily-kirichenko:
Yes, Lexer tests are still use it.

PR #1255, opened 2015-11-18
Fix the issue where Lint configs in nested folders weren't recognized

### LINKED ISSUE #1251: Lint settings on multi project solution (C#/F# etc)
Hi,

For some reason on my solution the F# Power Tools option for Lint says 'No projects are currently open'. When I create a new F# solution with just 1 project, the options appear fine. 

My current solution has C#, F# and SQL.

dungpa:
Read all F# projects in the solution recursively.

Fix #1251.

### DIFF
--- a/src/FSharpVSPowerTools.Logic/VSUtils.fs
+++ b/src/FSharpVSPowerTools.Logic/VSUtils.fs
@@ -6,11 +6,6 @@ open FSharpVSPowerTools
 
 
 type String with
-
-    /// Splits a string into lines for all platform's linebreaks.
-    /// If the string mixes windows, mac, and linux linebreaks, all will be respected
-    static member toLineArray str = String.getLines str
-
     /// Splits a string into lines for all platform's linebreaks.
     /// If the string mixes windows, mac, and linux linebreaks, all will be respected
     member self.ToLineArray () = String.getLines self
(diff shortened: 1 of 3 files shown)


src/FSharpVSPowerTools.Logic/VSUtils.fs
```diff
@@ -58,16 +53,16 @@ let isPhysicalFileKind (kind: string) =
     kind.Equals(EnvDTE.Constants.vsProjectItemKindPhysicalFile, StringComparison.OrdinalIgnoreCase)
 
 let isPhysicalFileOrFolderKind kind =
-    kind <> null && (isPhysicalFolderKind kind) || (isPhysicalFileKind kind)
+    isNotNull kind && (isPhysicalFolderKind kind) || (isPhysicalFileKind kind)
```
vasily-kirichenko: I don't think replacing all `= null` and `<> null` with `isNull` / `isNotNull` makes code better. These parts of code are called not very frequently and worse `=` `<>` performance should not drive code style here.

dungpa: I think `isNotNull` is as clear as `<> null`. I try to be consistent by using one approach and not to have to worry about whether the check will be in a tight loop.

vasily-kirichenko: I see. I think the perf of `=` `<>` must be fixed in the compiler. This situation is unhealthy :(

PR #1257, opened 2015-11-19
Printf specifiers tagger

vasily-kirichenko:
It works like this:

I'd like to use different color for this tags. Currently I reuse "MarkerFormatDefinition/HighlightedReference", as `HighlightUsageTagger` does. @dungpa do you know a way to add a color (visible in settings) and specify to use it in this new tagger? If I remember correctly, we did use a custom color for tags until we found out how to use the standard one.

### DIFF
--- a/src/FSharpVSPowerTools.Core/PrintfSpecifiersUsageGetter.fs
+++ b/src/FSharpVSPowerTools.Core/PrintfSpecifiersUsageGetter.fs
@@ -26,16 +26,19 @@ let getAll (input: ParseAndCheckResults): PrintfSpecifierUse[] option Async =
                 let ownSpecifiers, restSpecifiers = 
                     specifierRanges 
                     |> Array.partition (Range.rangeContainsRange func.String)
-            
-                if func.Args.Length > ownSpecifiers.Length then
-                    failwithf "Too many Printf arguments for %+A (%d > %d)" func func.Args.Length ownSpecifiers.Length
-
-                let uses = 
-                    func.Args
-                    |> Array.sortBy sortRange
-                    |> Array.zip (ownSpecifiers.[0..func.Args.Length - 1] |> Array.sortBy sortRange)
-                    |> Array.map (fun (specifier, arg) -> { SpecifierRange = specifier; ArgumentRange = arg })
-                restSpecifiers, uses :: acc
+                
+                match ownSpecifiers with
+                | [||] -> restSpecifiers, acc
+                | _ ->
+                    if func.Args.Length > ownSpecifiers.Length then
+                        failwithf "Too many Printf arguments for %+A (%d > %d)" func func.Args.Length ownSpecifiers.Length
+                    
+                    let uses = 
+                        func.Args
+                        |> Array.sortBy sortRange
+                        |> Array.zip (ownSpecifiers.[0..func.Args.Length - 1] |> Array.sortBy sortRange)
+                        |> Array.map (fun (specifier, arg) -> { SpecifierRange = specifier; ArgumentRange = arg })
+                    restSpecifiers, uses :: acc
                ) (specifierRanges, [])
             |> snd
             |> List.toArray
--- a/src/FSharpVSPowerTools.Core/PrintfSpecifiersUsageGetter.fs
+++ b/src/FSharpVSPowerTools.Core/PrintfSpecifiersUsageGetter.fs
@@ -27,14 +27,14 @@ let getAll (input: ParseAndCheckResults): PrintfSpecifierUse[] option Async =
                     specifierRanges 
                     |> Array.partition (Range.rangeContainsRange func.String)
             
-                if ownSpecifiers.Length > func.Args.Length then
-                    failwithf "Too many Printf specifiers for %+A (%d > %d)" func ownSpecifiers.Length func.Args.Length
+                if func.Args.Length > ownSpecifiers.Length then
+                    failwithf "Too many Printf arguments for %+A (%d > %d)" func func.Args.Length ownSpecifiers.Length
 
                 let uses = 
-                    ownSpecifiers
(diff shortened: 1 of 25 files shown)


dungpa:
Dig up the old code, here it is https://github.com/fsprojects/VisualFSharpPowerTools/blob/8c172d04050559655be75e28182ec8e3cc027641/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs#L21-L35. The main issue is to hook up theme change in order to have appropriate colors for each theme.

vasily-kirichenko:
I stole the color for light theme from R#:

image

vasily-kirichenko:
@dungpa what do you think we should use instead of these types (I've copy-pasted them from classifier) https://github.com/fsprojects/VisualFSharpPowerTools/blob/1339ba748a51c0529329096cccde9f17ee11dafb/src/FSharpVSPowerTools/Commands/PrintfSpecifiersUsageTaggerProvider.cs#L68-L72 ?

vasily-kirichenko:
Fixed the issue with themes. Now we should grab the proper Dark theme color from R#.

vasily-kirichenko:
@dungpa do you think we should add a separate sub-setting for this feature?

vasily-kirichenko:
I think it's ready.

vasily-kirichenko:
Wait, it does not work for user defined printf functions.

dungpa:
Here is the color from R# dark theme:
image

DarkThemeColor = Color.FromRgb(128, 255, 128);

vasily-kirichenko:
@dungpa I mean the tags color, not classifier color. I've grabbed it.

OK, finding printf argument positions is rather messy, especially if they are forward/backward piped.

That said, the feature works in most case.

src/FSharpVSPowerTools.Core/PrintfSpecifiersUsageGetter.fs
```diff
@@ -0,0 +1,46 @@
...
+
+open Microsoft.FSharp.Compiler
+open FSharpVSPowerTools.UntypedAstUtils
+
+[<NoComparison>]
+type PrintfSpecifierUse =
+    { SpecifierRange: Range.range
+      ArgumentRange: Range.range }
+
+let private sortRange (r: Range.range) = r.StartLine, r.StartColumn
```
dungpa: Does this mean 'startPos' i.e. a tuple denoting start position? (sortRange sounds like a totally different thing.)

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -1162,4 +1154,260 @@ module Outlining =
...
             let (ParsedImplFileInput (_, _, _, _, _, modules, _)) = implFile
             Seq.collect parseModuleOrNamespace modules
-        | _ -> Seq.empty
\ No newline at end of file
+        | _ -> Seq.empty
+
+module Printf =
+    [<NoComparison>]
+    type PrintfFunction = 
+        { String: Range.range
```
dungpa: Perhaps change `String` to `FormatString`?

src/FSharpVSPowerTools.Logic/PrintfSpecifiersUsageTagger.fs
```diff
@@ -0,0 +1,115 @@
...
+            return NormalizedSnapshotSpanCollection spans
+        } 
+        |> Async.bind (fun x ->
+            async {
+                spans <- x
+                return! callInUIContext <| fun _ -> tagsChanged.Trigger(self, SnapshotSpanEventArgs buffer.CurrentSnapshot.FullSpan)
+            }
+        )
+
+    let onCaretMoveListener: DocumentEventListener option ref = ref None
```
dungpa: Can we use `DocumentEventListener lazy` type instead? In the end, onCaretMoveListener is assigned at most once.

```fsharp
onCaretMoveListener := 
                               lazy(new DocumentEventListener ([ViewChange.layoutEvent view; ViewChange.caretEvent view], 200us, onCaretMove))
```

src/FSharpVSPowerTools/Commands/PrintfSpecifiersUsageTaggerProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    [Export]
+    public class PrintfColorManager
+    {
+        static readonly Color LightThemeColor = Color.FromRgb(245, 222, 179);
+        static readonly Color DarkThemeColor = Color.FromRgb(0, 77, 77);
+        VisualStudioTheme lastTheme = VisualStudioTheme.Unknown;
+
+        [Import]
+        ThemeManager themeManager = null;
```
dungpa: You can use `ImportingConstructor` here. It's better and more consistent with the rest of the project.

dungpa:
Looks great. I agree it should be a separate feature with its own setting.

vasily-kirichenko:
All done.

dungpa:
Awesome.

WildBamboo:
This super handy feature doesn't seem to work in fsx files for me, but works in fs files. Is this known and expected, or a bug?

dungpa:
@WildBamboo It's a bug. Could you open a separate with more information e.g. (1) What is location of FSharp.Core 4.3.1.0 in your system (2) Can you see any error logged in Diagnostics mode http://fsprojects.github.io/VisualFSharpPowerTools/faq.html#What-is-Diagnostic-mode-When-should-I-use-it? Thanks.

WildBamboo:
Created issue 1257 and in the process got some more info. It appears `<|` usage breaks things. Probably what @vasily-kirichenko was getting at above:

> OK, finding printf argument positions is rather messy, especially if they are forward/backward piped.

Issue #1256, opened 2015-11-19
[Outlining] ObjectDisposedException in log


vasily-kirichenko:
```
[VFPT][Error 04:06:04 ] Exception Message: Cannot access a disposed object.
Object name: 'OutliningManager'.
Stack Trace:    at Microsoft.VisualStudio.Text.Outlining.OutliningManager.EnsureValid()
   at Microsoft.VisualStudio.Text.Outlining.OutliningManager.EnsureValid(SnapshotSpan span)
   at Microsoft.VisualStudio.Text.Outlining.OutliningManager.CollapseAll(SnapshotSpan span, Predicate`1 match)
   at FSharpVSPowerTools.OutliningTaggerProvider.<>c__DisplayClass7.<TextViewCreated>b__4(Object sender, SnapshotSpanEventArgs e) in d:\github\FSharpVSPowerTools\src\FSharpVSPowerTools\Commands\OutliningTaggerProvider.cs:line 90
   at Microsoft.FSharp.Control.FSharpEvent`2.Trigger(Object sender, TArgs args)
   at FSharpVSPowerTools.Outlining.OutliningTagger.triggerUpdate(ScopeSpan[] newSnapshotSpans) in D:\github\FSharpVSPowerTools\src\FSharpVSPowerTools.Logic\Outlining\OutliningTagger.fs:line 90
   at FSharpVSPowerTools.Outlining.clo@178-495.Invoke(Unit _arg1) in D:\github\FSharpVSPowerTools\src\FSharpVSPowerTools.Logic\Outlining\OutliningTagger.fs:line 178
   at FSharpVSPowerTools.ProjectSystem.VSUtils.protectOrDefault[a](FSharpFunc`2 f, a defaultVal) in D:\github\FSharpVSPowerTools\src\FSharpVSPowerTools.Logic\VSUtils.fs:line 431
```

dungpa:
Is this with current master? Do you have a reliable way to reproduce this?

vasily-kirichenko:
yes, it's with current master. I happened on just after a file was opened. I'll try to make minimal repro.

dungpa:
Does this still occur? It might not happen any more due to #1260.

vasily-kirichenko:
I have not checked but it's definetely fixed.

vasily-kirichenko:
Master, VS 2015 Update 1:

```
[VFPT][Error 11:24:41 AM] Exception Message: Cannot access a disposed object.
[...]

PR #1260, opened 2015-11-21
Prevent Dispose being called in outlining tagger at wrong time

dkinitiate:
This is some serious Voodoo business.

One time you open a view with a file, and everything works well. `Dispose` is called on the Outlining Tagger as hoped when the view is closed.

The second time?
`Dispose` is called.
When?
After we have calculated what to outline, and alerted VS that there are tags that it needs to show.

Here's a callstack of said event:
screenshot 26

The fix? Removing the `IDisposable` imeplementation from the tagger. Not ideal, obviously, but it seems to solve this bug.

### DIFF
--- a/src/FSharpVSPowerTools.Logic/Outlining/OutliningTagger.fs
+++ b/src/FSharpVSPowerTools.Logic/Outlining/OutliningTagger.fs
@@ -446,6 +446,9 @@ type OutliningTagger
             |> Seq.filter (fun s -> normalizedSnapshotSpans.IntersectsWith s.SnapSpan)
             |> Seq.choose createTagSpan
 
+    // prevent compilation error, docEventListener must have some code that 
+    // references it.
+    member private x.DocEventListener = docEventListener
 
     interface ITagger<IOutliningRegionTag> with
         member __.GetTags spans =
@@ -453,9 +456,3 @@ type OutliningTagger
 
         [<CLIEvent>]
         member __.TagsChanged = tagsChanged.Publish
-
-
-    interface IDisposable with
-        member __.Dispose () =
-            docEventListener.Dispose ()
-            scopedSnapSpans <- [||]
--- a/src/FSharpVSPowerTools.Logic/Outlining/OutliningTagger.fs
+++ b/src/FSharpVSPowerTools.Logic/Outlining/OutliningTagger.fs
@@ -180,7 +180,7 @@ type OutliningTagger
         |> Async.Ignore
 
     /// viewUpdate -=> doUpdate -=> triggerUpdate -=> tagsChanged
-    let docEventListener =
+    let _docEventListener =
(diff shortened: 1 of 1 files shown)


vasily-kirichenko:
I don't think it's a good idea to not dispose `DocumentEventListener`, it may cause `OutlineTagger` to leak (or not?) 

I don't understand what concrete `TagAggregator` is gathering the tags and why it disposes our tagger.

I set a breakpoint in `Dispose` and did not see it being called in any time other than when a document is closing (in both VS 2015 and 2013).

dkinitiate:
I agree that it's not a good idea. but then again it's the only workaround I could come up with on short notice.

The downside to this approach is simply that the DocEventListener is still alive for a bit - after the view is closed, and until the GC collects it. Not ideal yes, but I can't think of a way that the buffer can change while the view is closed either, so I think overall this is a pretty cheap workaround.

Also, this is something that I managed to reproduce consistently, using the reproduction method you showed me - open the VFPT solution and close/open a file repeatedly. Every odd time this is done (no outlining):

- the tagger is created.
- it calculates spans.
- it triggers the event VS listens to
- VS calls `Dispose` on it (?!?@!#)
- When it asks it for tags, the `Dispose` method makes sure there were no spans to report.
- The tagger doesn't do any more work as the event listener has been disposed

dungpa:
I think it's a behavior change (bug?) in VS 2015. The tag aggregator also calls `CreateTagger` and shares the same tagger with us (we use `GetOrCreateSingletonProperty` there). When they're done, they just dispose our tagger and outlining stops working.

Another way to fix this is to always create a fresh tagger in `CreateTagger`. VS throws away a tagger very soon so it's not much less efficient in doing so. What do you think?

dkinitiate:
I mused about that exact other fix that you've just proposed over at Slack, yes.

Unfortunately it's only a theory that this approach would work at the moment. Even if that solves it however, there is still the issue of `public void TextViewCreated(IWpfTextView textView)` relying on the tagger existing in the `buffer` properties.

dungpa:
Oh well, you're right.

vasily-kirichenko:
I'm for creating fresh tagger each time anything calls `CreateTagger` and revert `Dispose`.

dungpa:
It requires duplicating the work at least twice for collapsing outlining by default.

I'm for removing collapsing by default. It doesn't work very well (collapsing some time after files have been opened). And it causes this issue and #1256, which I have no idea how to fix.

We should really look into saving outlining to `suo` files, which helps persist collapsing states across sessions /cc @cloudRoutine.

vasily-kirichenko:
I disagree. I love auto collapsing. Actually, it's the only case where I use Outlining at all - auto collapse opens and # stuff.

dungpa:
I'm fine with either way: removing GetOrCreate or removing Dispose. Please proceed.

PR #1261, opened 2015-11-22
Improve theme handling

### LINKED ISSUE #990: Syntax Coloring reverts to default on Hibernate/Sleep/Remote Access
I think the only reason I noticed this was because a regularly work on a laptop and and I have a dark color scheme with highly customized colors.

If the machine goes to sleep, hibernates, or if there is a remote desktop connection (occurs both when the remote connection is initiated, and when you log back in to the machine after a remote connection) the colors appear to revert to their default settings, which in my case means I can't see a lot of the code. This doesn't correct itself until I close/reopen visual studio.

I noticed this with VS2012 and VS2013. I've not tried VS2015 yet.

Let me know if there is any additional info that would be useful tracking this down.

dungpa:
After a RDP connection, Windows may initiate a system theme change event (see http://stackoverflow.com/questions/33064295/tlistview-columns-order-bug-after-windows-theme-change). Because we haven't saved a current theme, we may blindly update themes when it's not really needed. This PR added extra checks to fix the issue.

Close #990.


src/FSharpVSPowerTools.Logic/ThemeManager.fs
```diff
@@ -62,14 +63,11 @@ type ThemeManager [<ImportingConstructor>]
...
-                let dte = serviceProvider.GetService<DTE, SDTE>()
-                let fontsAndColors = 
-                    // This method will not work well for customized themes, but we should not give up.
-                    dte.Properties("FontsAndColors", "TextEditor").Item("FontsAndColorsItems").Object :?> FontsAndColorsItems
-                let background = ColorTranslator.FromOle(int (fontsAndColors.Item("Plain Text").Background))
-                match background.R, background.G, background.B with
-                | 30uy, 30uy, 30uy -> VisualStudioTheme.Dark
-                | _ -> VisualStudioTheme.Light
+                let color = VSColorTheme.GetThemedColor EnvironmentColors.ToolWindowTextColorKey
+                if color.GetBrightness() > 0.5f then
```
vasily-kirichenko: wow :)

src/FSharpVSPowerTools/Commands/PrintfSpecifiersUsageTaggerProvider.cs
```diff
@@ -85,20 +85,22 @@ public PrintfColorManager(ThemeManager themeManager, IEditorFormatMapService edi
...
+            return _currentTheme == VisualStudioTheme.Dark ? DarkThemeColor : LightThemeColor;
         }
 
         public void UpdateColors()
         {
-            var currentTheme = _themeManager.GetCurrentTheme();
+            var newTheme = _themeManager.GetCurrentTheme();
 
-            if (currentTheme != VisualStudioTheme.Unknown && currentTheme != lastTheme)
+            if (newTheme != VisualStudioTheme.Unknown && newTheme != _currentTheme)
```
vasily-kirichenko: This logic is duplicated in `SyntaxConstructClassifierProvider`. Maybe move it to `VSColorTheme` so that `ThemeChanged` event is not raised if theme has not actually changed?

src/FSharpVSPowerTools.Logic/ThemeManager.fs
dungpa: It's hackish but it works for custom themes e.g. Solarized (Dark).

src/FSharpVSPowerTools/Commands/PrintfSpecifiersUsageTaggerProvider.cs
dungpa: VSColorTheme is a built-in class in VS SDK. I'm not sure how to reduce duplication here.

vasily-kirichenko: ok

vasily-kirichenko:
will test it in the rdp scenario tomorrow.

PR #1264, opened 2015-11-24
Outline comments

vasily-kirichenko:

### DIFF
--- a/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
@@ -800,6 +800,7 @@ module Outlining =
         | RecordDefn = 41
         | UnionDefn = 42
         | Comment = 43
+        | XmlDocComment = 44
 
     [<NoComparison; Struct>]
     type ScopeRange (scope:Scope, collapse:Collapse, r:range) =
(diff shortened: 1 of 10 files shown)


vasily-kirichenko:
"Comments enabled" setting works OK, but "Comments enabled" does not: open the dialog, uncheck it, close the dialog, open again - it's checked :( No idea why. Checked everything 5 times.

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -1149,11 +1150,58 @@ module Outlining =
...
+    type private LineNum = int
+    type private LineStr = string
+    type private MultilineComment = (LineNum * LineStr) list
+
+    let getCommentRanges (lines: string[]) =
+        let comments: MultilineComment list =
+            lines
+            |> Array.foldi (fun ((lastLineNum, currentComment, comments) as state) lineNum lineStr ->
+                match lineStr.TrimStart() with
+                | x when x.StartsWith "//" ->
```
dungpa: We have `StartsWith` active patterns in Core project.

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -1149,11 +1150,58 @@ module Outlining =
               yield! collectOpens decls
               yield! Seq.collect parseDeclaration decls }
 
-    let getOutliningRanges tree =
+    type private LineNum = int
+    type private LineStr = string
+    type private MultilineComment = (LineNum * LineStr) list
```
dungpa: Is this really `MultilineComment` (comments between (* and *) )? I think it's more like `CommentList`.

src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
```diff
@@ -1149,11 +1150,58 @@ module Outlining =
...
+                | _ -> lastComment :: comments
+                |> List.rev
+
+        comments
+        |> List.choose (fun comment ->
+            match comment with
+            | [] | [_] -> None
+            | _ ->
+                let comment = comment |> List.rev |> Array.ofList
+                let startLine, endLine = fst comment.[0], fst comment.[comment.Length - 1]
```
dungpa: Maybe reorganize these a bit:

```fsharp
let startLine, startStr = comment.[0]
let endLine, endStr = comment.[comment.Length - 1]
let startCol = startStr.IndexOf '/'
let endCol = endStr.TrimEnd().Length
```

dungpa:
Could you add outlining tests that contain comments? 

This PR doesn't recognize comments in a precise way e.g. having false positives on `//` inside multi-line strings and `//` inside block comments, but I think it's a very good start.

vasily-kirichenko:
All done + a distinct category for XML doc comments.

But the bug with storing "Enabled" setting for the two new categories is still here - it cannot be set to "false": uncheck the checkbox, close the dialog, open it again, the checkbox is checked. This is reproduced for only two checkboxes - [x] Enable comments and [x] Enable XML doc comments. I've checked all the UI code again, twice, no results :(

dungpa:
Should `_control.Comments.Enabled` be `_control.Comments.Enabled` in https://github.com/vasily-kirichenko/FSharpVSPowerTools/blob/b507b7ef338670cc141ba8dc06f3f4cebbfc1f4f/src/FSharpVSPowerTools/UI/OutliningOptionsPage.cs#L99-L105? The `Enabled` property is inherited from somewhere and it has a different purpose.

vasily-kirichenko:
OMG, yes! Tested all combinations of comments settings - all works. Thanks a lot.

dungpa:
:+1:

PR #1269, opened 2015-11-28
Disable Linting by default

### LINKED ISSUE #1266: Please turn off end-of-line whitespace warnings by default
I'm spending some time working with a finance industry customer. I just upgraded VFPT in that context and the "end of line whitespace warnings" came on by default. I found this felt wrong: I've personally never cared about end-of-line whitespace in 30 years of programming, so for me these warnings are spurious, and that's likely to be the case for a substantial number of other people - except where the majority of people on a project are using specific editors that (mistakenly IMHO) remove whitespace by default.

So I recommend this warning be off by default (or FSharpLint should be off by default, which I've suggested for possible performance reasons). It seems wrong for beginners (or even experienced F# programmers) to be asked to have an opinion about this setting and to then have to dig into options and disable something. The warning squiggles are intrusive enough that the substantial number of people who don't like the setting on by default will have to go turn it off explicitly. And it seems wrong that I find myself turning off FSharpLint just to get rid of these warnings (I know I don't have to turn off FSharpLint, but it's the quickest way to make the warnings go away).

thx
Don

dungpa:
It only affects new users that don't have VFPT installed before.

I hope we can enable it again with typecheck results when performance issues are sorted out.

Close #1266.


vasily-kirichenko:
:+1:

Issue #1274, opened 2015-12-01
Backwards pipe breaks printf specifiers tagger


WildBamboo:
`C:\Program Files (x86)\Reference Assemblies\Microsoft\FSharp\.NETFramework\v4.0\4.3.1.0\FSharp.Core.dll`

When opening a fsx file I see the following diagnostics:

```
[VFPT][Error 02:32:15 p.m.] Context: Failed to update printf specifier usages.
Exception Message: Too many Printf arguments for {FormatString =
  C:\...\tools.fsx (349,37--349,80) IsSynthetic=false;
 Args =
  [|C:\...\tools.fsx (349,81--349,88) IsSynthetic=false;
    C:\...\tools.fsx (349,12--349,25) IsSynthetic=false|];} (2 > 1)
Stack Trace:    at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll@34-4.Invoke(String message)
   at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll$cont@33(FSharpList`1 acc, PrintfFunction func, range[] restSpecifiers, range[] ownSpecifiers, Unit unitVar)
   at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll@25-2.Invoke(PrintfFunction func)
   at Microsoft.FSharp.Collections.ArrayModule.Fold[T,TState](FSharpFunc`2 folder, TState state, T[] array)
   at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll@16.Invoke(range[] _arg1)
   at Microsoft.FSharp.Control.AsyncBuilderImpl.callA@803.Invoke(AsyncParams`1 args)
[VFPT][Information 02:32:15 p.m.] [SyntaxConstructClassifier] [Fast stage] 
Total = 00:00:00.4980844
getCategoriesAndLocations, 00:00:00.1952084
ParseFileInProject, 00:00:00.1674228
LS GetAllUsesOfAllSymbolsInFile, 00:00:00.0028948
LS allSymbolsUses, 00:00:00.0013133
LS ParseAndCheckFileInProject, 00:00:00.0003939

[VFPT][Information 02:32:15 p.m.] [SyntaxConstructClassifier] Replace spans entirely because new spans has no errors.
[VFPT][Information 02:32:15 p.m.] [SyntaxConstructClassifier] [Slow stage] 
Total = 00:00:00.5231186
[...]

PR #1282, opened 2015-12-04
Do not throw all calculated Printf argument highlights if some of them are erroneous

vasily-kirichenko:
It does not fix this https://github.com/fsprojects/VisualFSharpPowerTools/issues/1274 and other possible AST forms on which the feature does not work, but it prevents loosing all Printf highlight tags to disappear if a single such error occurs.


dungpa:
:+1:

PR #1286, opened 2015-12-06
Peek definition

vasily-kirichenko:


vasily-kirichenko:
- [ ] Add a setting
- [ ] Should it show metadata for external symbols?

dungpa:
My answer is yes for all questions. How do you solve the 3rd issue?

vasily-kirichenko:
I' not sure the third one should be implemented. The nice thing about the current impl is that it opens instantly, compared to 2+ seconds for ordinar file opening. I'm afraid that if we make it a full fledged buffer, it will become equally slow.

dungpa:
Ok, let's finish the first two items. The last one should only be done if file opening is fast enough.

vasily-kirichenko:
I'm not sure I'm ready to port Go to metadata code, it's tightly coupled with UI in `GoToDefinition.fs`. I'd rather leave it to implement in the future.

dungpa:
Fair enough.

The new Peek API isn't available on VS 2013. I've got this error when invoking Alt + F12 on VS 2013.

> <description>System.ComponentModel.Composition.CompositionException: The composition produced a single composition error, with 2 root causes. The root causes are provided below. Review the CompositionException.Errors property for more detailed information.&#x000D;&#x000A;&#x000D;&#x000A;1) The export &apos;Microsoft.VisualStudio.Editor.Implementation.PeekResultFactory (ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekResultFactory&quot;)&apos; is not assignable to type &apos;Microsoft.VisualStudio.Language.Intellisense.IPeekResultFactory&apos;.&#x000D;&#x000A;&#x000D;&#x000A;Resulting in: Cannot set import &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider..ctor (Parameter=&quot;peekResultFactory&quot;, ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekResultFactory&quot;)&apos; on part &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&apos;.&#x000D;&#x000A;Element: FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider..ctor (Parameter=&quot;peekResultFactory&quot;, ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekResultFactory&quot;) --&gt;  FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&#x000D;&#x000A;&#x000D;&#x000A;Resulting in: Cannot get export &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider (ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekableItemSourceProvider&quot;)&apos; from part &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&apos;.&#x000D;&#x000A;Element: FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider (ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekableItemSourceProvider&quot;) --&gt;  FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&#x000D;&#x000A;&#x000D;&#x000A;2) The export &apos;Microsoft.VisualStudio.Text.Implementation.TextDocumentFactoryService (ContractName=&quot;Microsoft.VisualStudio.Text.ITextDocumentFactoryService&quot;)&apos; is not assignable to type &apos;Microsoft.VisualStudio.Text.ITextDocumentFactoryService&apos;.&#x000D;&#x000A;&#x000D;&#x000A;Resulting in: Cannot set import &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider..ctor (Parameter=&quot;textDocumentFactoryService&quot;, ContractName=&quot;Microsoft.VisualStudio.Text.ITextDocumentFactoryService&quot;)&apos; on part &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&apos;.&#x000D;&#x000A;Element: FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider..ctor (Parameter=&quot;textDocumentFactoryService&quot;, ContractName=&quot;Microsoft.VisualStudio.Text.ITextDocumentFactoryService&quot;) --&gt;  FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&#x000D;&#x000A;&#x000D;&#x000A;Resulting in: Cannot get export &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider (ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekableItemSourceProvider&quot;)&apos; from part &apos;FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&apos;.&#x000D;&#x000A;Element: FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider (ContractName=&quot;Microsoft.VisualStudio.Language.Intellisense.IPeekableItemSourceProvider&quot;) --&gt;  FSharpVSPowerTools.Logic.VS2015.PeekableItemSourceProvider&#x000D;&#x000A;&#x000D;&#x000A;   at System.ComponentModel.Composition.Hosting.CompositionServices.GetExportedValueFromComposedPart(ImportEngine engine, ComposablePart part, ExportDefinition definition)&#x000D;&#x000A;   at System.ComponentModel.Composition.Hosting.CatalogExportProvider.GetExportedValue(CatalogPart part, ExportDefinition export, Boolean isSharedPart)&#x000D;&#x000A;   at System.ComponentModel.Composition.Hosting.CatalogExportProvider.CatalogExport.GetExportedValueCore()&#x000D;&#x000A;   at System.ComponentModel.Composition.Primitives.Export.get_Value()&#x000D;&#x000A;   at System.ComponentModel.Composition.ExportServices.GetCastedExportedValue[T](Export export)&#x000D;&#x000A;   at System.ComponentModel.Composition.ExportServices.&lt;&gt;c__DisplayClass4`2.&lt;CreateStronglyTypedLazyOfTM&gt;b__1()&#x000D;&#x000A;   at System.Lazy`1.CreateValue()&#x000D;&#x000A;   at System.Lazy`1.LazyInitValue()&#x000D;&#x000A;   at System.Lazy`1.get_Value()&#x000D;&#x000A;   at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.InvokeMatchingFactories[TExtensionInstance,TExtensionFactory,TMetadataView](IEnumerable`1 lazyFactories, Func`2 getter, IContentType dataContentType, Object errorSource)</description>

Could we only enable the feature for VS2015+? It's even better to show it clearly in the setting e.g. `Peek Definition (VS2015+)`.

vasily-kirichenko:
I know, that is why I put it into Logic.VS2015 project. Maybe just hide or make "disabled" (gray and non interactive) the setting in VS2013?

dungpa:
We should make sure that Alt+F12 on VS2013 should do nothing. A clear label is OK. Of course, graying out the setting on VS 2013 is better :-).

vasily-kirichenko:
Huh. I try to fit this trick https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/NavigateToItem.fs#L277-L295, but the case is not the same because there's no `IPeekableItemSourceProvider` in VS2013 at all, so we cannot add a fake implementation in Logic, then select right one at runtime. Any ideas?

dungpa:
Try this first https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools/Commands/UnionPatternMatchCaseGeneratorSmartTaggerProvider.cs#L55-L57. It might be enough.

vasily-kirichenko:
Done both disabling the setting and the trick with returning `null` (your last comment). 
In VS 2015 everything works.

vasily-kirichenko:
About the bug, it seems only Highlight refs stops working after single Peek Definitions invoked:

vasily-kirichenko:
I think I understand why the bug: peek definition seems to reuse already open buffers and, when I close peek def window, it's `Dispose` is called, which disposes all our timers => most features stop working.

So, it seems we should use `ITextView.Properties.GetOrCreateSingletonProperty` instead of `ITextBuffer.GetOrCreateSingletonProperty`. Will check shortly.

vasily-kirichenko:
No, it has not helped :(

vasily-kirichenko:
Oh, no https://github.com/dotnet/roslyn/blob/a4e375b95953e471660e9686a46893c97db70b0e/src/EditorFeatures/Core/Shared/Extensions/ITextViewExtensions.PerSubjectBufferProperty.cs

vasily-kirichenko:
It will be a long story.

dungpa:
The link looks scary :(

It throws errors on VS2013; we need another way.

vasily-kirichenko:
Have fixed the bug with stopping all the features working after opening-closing a Peek Def view.

vasily-kirichenko:
I have no idea how to get rid of the exception in VS 2013. I suggest:

* stop supporting it
* create different VSIXs for 2013 and 2015+

src/FSharpVSPowerTools/UI/GeneralOptionsControl.cs
```diff
@@ -181,6 +187,13 @@ private void GeneralOptionsControl_Load(object sender, EventArgs e)
             QuickInfoPanelEnabled = _optionsPage.QuickInfoPanelEnabled;
             LinterEnabled = _optionsPage.LinterEnabled;
             OutliningEnabled = _optionsPage.OutliningEnabled;
+            PeekDefinitionEnabled = _optionsPage.PeekDefinitionEnabled;
+
+            if (!_optionsPage.PeekDefinitionAvailable)
+            {
+                chbPeekDefinition.Enabled = false;
+                chbPeekDefinition.Text = chbPeekDefinition.Text + " (VS2015+ only)";
```
dungpa: This will keep adding VS2015+ suffix to the label each time we open General options dialog.

src/FSharpVSPowerTools.Logic.VS2015/PeekDefinition.fs
```diff
@@ -0,0 +1,136 @@
...
+
+    interface IPeekableItemSource with
+        member __.AugmentPeekSession (session, peekableItems) =
+            asyncMaybe {
+                do! if String.Equals(session.RelationshipName, PredefinedPeekRelationships.Definitions.Name, 
+                                     StringComparison.OrdinalIgnoreCase) then Some() else None
+                let! triggerPoint = session.GetTriggerPoint(buffer.CurrentSnapshot) |> Option.ofNullable
+                let! symbolSpan, definitionRange = getDefinitionRange triggerPoint
+                peekableItems.Add (DefinitionPeekableItem(symbolSpan, definitionRange, peekResultFactory))
+            } |> Async.RunSynchronously |> ignore
```
dungpa: Could we use `Async.StartInThreadPoolSafe` or `Async.StartImmediateSafe` here instead?

dungpa:
Let's merge this after addressing the last two comments. I'll try to disable the feature for VS2013 after merging the PR.

src/FSharpVSPowerTools/UI/GeneralOptionsControl.cs
vasily-kirichenko: Yeah, I suspected it.

src/FSharpVSPowerTools.Logic.VS2015/PeekDefinition.fs
vasily-kirichenko: I tried. It just silently stopped working.

dungpa: OK, synchronous API then :(

vasily-kirichenko:
I think we should improve it significantly:

* To fix the issue with `OpenDocumentTracker` it should keep number of views opened for concrete buffer and countdown it when a view is closing. When the counter is zero, remove it from opened docs map.
* The root cause of the fact that no features work inside Peek Def view is that `DTE` returns main document as active over here https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/VSUtils.fs#L252, then this case does not match if Peek view contains document different to the one opened in the main view https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/VSUtils.fs#L262

dungpa:
Yes, we should eventually get rid of ActiveDocument thing. It has never been working well.

PR #1290, opened 2015-12-08
Make sure that Peek Definition menu items are disabled on VS2013

dungpa:
Could you check it should not affect VS2015? I'll double check the behaviour on VS2013 tomorrow.


dungpa:
No peek definition (as expected) in VS2013.

PR #1291, opened 2015-12-09
Upgrade FSharp.Core to 4.3.1.0

cloudRoutine:


ReedCopsey:
This would be very nice, as it'd allow us to use PCL versions of libraries instead of forcing 4.3.0.0 compatibility.

vasily-kirichenko:
I've tested it in VS 2015, works OK. Although I'm sure it works in VS 2013, too, @dungpa, can you test it in it please?

dungpa:
Works fine on VS2013.

PR #1294, opened 2015-12-10
WIP - Revamp VFPT Options/Settings [Extenal Xml Storage] 

cloudRoutine:
Looking for some feedback/review on the rough implementation of the new settings system.

The settings are saved/loaded from an xml file in `%USER%\.configs\.vfpt\vfpt_settings.xml`

The VsPackage no longer provides services for the different kinds of settings.

I know there's plenty of room for refinement (the interfaces are probably unnecessary, the settings store does a lot of unnecessary reads and writes, codegen is only implemented partially, etc) I'll get to it after I've gotten some feedback.


cloudRoutine:
[Settings.fs](https://github.com/cloudRoutine/VisualFSharpPowerTools/blob/xml-settings/src/FSharpVSPowerTools.Logic/Setting.fs) is where almost all the changes are.
@vasily-kirichenko @dungpa

src/FSharpVSPowerTools.Logic/FSharpVSPowerTools.Logic.fsproj
```diff
@@ -35,11 +35,11 @@
     <Tailcalls>true</Tailcalls>
     <OutputPath>bin\Release\</OutputPath>
     <DefineConstants>TRACE</DefineConstants>
-    <WarningLevel>5</WarningLevel>
+    <WarningLevel>3</WarningLevel>
     <DocumentationFile>bin\Release\FSharpVSPowerTools.Logic.XML</DocumentationFile>
     <OtherFlags>--warnon:1182</OtherFlags>
     <NoWarn>52</NoWarn>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
vasily-kirichenko: revert these settings

src/FSharpVSPowerTools.Logic/Setting.fs
```diff
@@ -1,42 +1,470 @@
...
+
+        member this.GetOrCreateElement name =
+            if this.HasElement name then this.Element name else this.AddElement name
+
+        member this.SetOrCreateElement name value =
+            if this.HasElement name then (this.Element name).Value <- value 
+            else (this.AddElement name).Value <- value
+
+
+    type Dictionary<'key,'value> with
```
vasily-kirichenko: we already have some Dictionaty related functions here https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Core/Utils.fs#L735

src/FSharpVSPowerTools.Logic/FSharpVSPowerTools.Logic.fsproj
cloudRoutine: I'll change it back once this feature is ready to merge. Having it on really annoys me as I work.

src/FSharpVSPowerTools.Logic/Setting.fs
cloudRoutine: Should `AddOrUpdate` be changed to a module function of `Dict` or should both type extensions be kept and moved to `Utils.fs`?  
The `XContainer` type extensions be moved there too, right?

src/FSharpVSPowerTools.Logic/FSharpVSPowerTools.Logic.fsproj
vasily-kirichenko: How about add a new light bomb to solve it: "Annotate with NoComparison attribute", "Annotate with NoEquality attribute"? :)

src/FSharpVSPowerTools.Logic/Setting.fs
vasily-kirichenko: Yes, I think module fucntions are better since they are pipable. 

Yes, `XContainer` extensions should also be put into Core.

dungpa:
It seems complex to replace a global setting system to another global setting system? What's the main motivation here?

vasily-kirichenko:
+1 for @dungpa 's comment. I don't see why it's an improvement.

cloudRoutine:
@dungpa Some benefits :
- In the process of doing dev work on and using VFPT my settings are constantly getting reset and modified without any input from me _(on multiple occassions I've seen the syntax highlighting and outlining in the main VS instance change while I'm testing something in the Exp instance)_, a more static approach like this affords protection from that.
- External settings can be exported, imported and shared with other people
- They make it easier to enable multiple configuration profiles than the current system
- People using both vs2013 and vs2015 can use a common set of configuration settings
- Visual Studio does not synchronize extension settings, this makes it easier to have a consistent config across machines and keep settings after reinstalls
- Could facilitate different settings for us when running VFPT in an experimental instance
- Easier to extend and modify than the current system, especially for new contributors.
- Cleaner and more straightforward implementation, no need to jump back and forth between the C# and F# projects to modify/work with configuration settings.
- Potential for settings that can work across editors (VS,XS,Ionide-Atom,Ionide-Vscode,etc.)
- Makes it easier to implement a set of global linting settings that can be superceded by the solution/repo specific settings of the current approach.

The way the settings are currently setup are oddly coupled and difficult to work with when adding a new feature. Each setting is defined as a service provided by the VFPT VsPackage that the package creates from the implementation of an interface in a dialog page that needs to be accessed in the Logic project using a service provider. Which means I need to implement a new GUI element(s) just to have settings for a new feature? I find this excessively convoluted and unnecessary.

> [A service is a contract between two VSPackages. One VSPackage provides a specific set of interfaces for another VSPackage to consume. Visual Studio is itself a collection of VSPackages that provides services to other VSPackages.](https://msdn.microsoft.com/en-us/library/bb166804.aspx)

It doesn't make sense that we're using a serivce to provide the settings for a package to itself.

Not to mention when the settings and their defaults are tightly coupled to the implementation of a GUI element it makes changing or redesigning the GUI element slightly more difficult and extremely more tedious to implement.

[`DesignerSerializationVisibilityAttribute `](https://msdn.microsoft.com/en-us/library/system.componentmodel.designerserializationvisibilityattribute.aspx#Anchor_7 ) :
> When a serializer persists the persistable state of a design mode document, it often adds code to the initialization method of components to persist values of properties that have been set at design time. This happens by default for most basic types, if no attribute has been set to direct other behavior.
> 
With the DesignerSerializationVisibilityAttribute, you can indicate whether the value for a property is Visible, and should be persisted in initialization code, Hidden, and should not be persisted in initialization code, or consists of Content, which should have initialization code generated for each public, not hidden property of the object assigned to the property

^ I don't understand why we're using this in the dialog pages.

This approach can be simlified even further by getting rid of the MEF component export entirely and just using the instantiations in the `SettingsContext` module.
( I tried to do it with static types, but I kept running into type initialization errors. I'm going to keep trying to get that to work )

Rather than being complex, I think this is a simplification

dungpa:
I agree that this solution is more flexible. Could you drive it into mergeable state? Particularly it should be compatible with our VS unit tests.

cloudRoutine:
the process of rebasing or possibly some other changes I made have created some intractable issues getting the test suite to pass. Going to try again on a new branch off of the current master

PR #1304, opened 2015-12-30
Fix for bug #1302

### LINKED ISSUE #1302: [Highlight printf] Exception
```
[VFPT][Error 10:35:39 AM] Context: Failed to update printf specifier usages.
Exception Message: Index was outside the bounds of the array.
Stack Trace:    at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll$cont@33(FSharpFunc`2 onError, FSharpList`1 acc, PrintfFunction func, range[] restSpecifiers, range[] ownSpecifiers, Unit unitVar)
   at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll@25-2.Invoke(PrintfFunction func)
   at Microsoft.FSharp.Collections.ArrayModule.Fold[T,TState](FSharpFunc`2 folder, TState state, T[] array)
   at FSharpVSPowerTools.PrintfSpecifiersUsageGetter.getAll@16.Invoke(range[] _arg1)
   at Microsoft.FSharp.Control.AsyncBuilderImpl.callA@851.Invoke(AsyncParams`1 args)
```

shanecharles:
Resolves #1302. No longer add the individual argument when the expression is an equality operation when evaluating `Printf.getAll`. Added some other tests which identified an issue of `Printf.getAll` not returning any functions when the equality operator is present.

### DIFF
--- a/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/UntypedAstUtils.fs
@@ -1324,6 +1324,12 @@ module Printf =
                     result.Add res
                 | _ -> ()
                 appStack := []
+            | SynExpr.App (_, _, SynExpr.App(_, true, SynExpr.Ident op, e1, _), e2, _) ->
+                addAppWithArg { Range = e.Range; Arg = e2.Range }
+                if op.idText <> Microsoft.FSharp.Compiler.PrettyNaming.opNameEquals then
+                    addAppWithArg { Range = e.Range; Arg = e1.Range }
+                walkExpr e2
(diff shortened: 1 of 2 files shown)


dungpa:
Thanks for the bug fix.

It looks good. But I think the logic should also be applied for any infix operator, not only `=` operator.

For example, this test case still fails:

```fsharp
System.Exception(("" > sprintf "%d" 1).ToString())
```

Would you mind updating the PR to handle it?

shanecharles:
Sure. I can make that change.

dungpa:
Great jobs. Thanks a lot.

PR #1309, opened 2016-01-02
Prioritize the args on the right side of printf statement.

### LINKED ISSUE #1308: Printf tagger highlights incorrectly when args appear on both sides
If there are args appearing on both sides of a `printf` the highlighting goes by position left to right. 
right-hand-arg-with-pipe

shanecharles:
Resolves #1308. Partitioning the args into left and right side, sorting each side, and combining them right side first. (With tests)
prioritize-args-right-of


vasily-kirichenko:
Great job again :) Thanks.

PR #1311, opened 2016-01-04
update packages

vasily-kirichenko:
pin NUnit to ~> 2
fix generate.fsx


vasily-kirichenko:
I have no time to fight with all these half backed packages issues.

PR #1322, opened 2016-01-15
Fix curried functions with lightweight interfaces.

### LINKED ISSUE #1315: Lightweight interface implementation does not work for members with curried arguments
If a member of an interface has arguments that contain multiple partially arguments, lightweight interface definitions don't work e.g. Suave's Logger interface looks like this in the heavyweight mode: -

```fsharp
member __.Log(arg1: LogLevel) (arg2: unit -> LogLine): unit = 
    failwith "Not implemented yet"
```

but the lightweight one looks like this: -
```fsharp
member __.Log(arg1 arg2) = failwith "Not implemented yet"
```
it should probably look like this instead

```fsharp
member __.Log arg1 arg2 = failwith "Not implemented yet"
```

isaacabraham:
Fixed #1315.


dungpa:
Look great. Thanks.

Issue #1335, opened 2016-02-07
printf slot highlighting doesn't handle * slots correctly


latkin:
Format specifiers with `*` consume 2 arguments, but the highlighting currently maps the specifier only to the first arg (and only the first arg to the specifier - the second arg gets incorrect mapping).

![printf](https://cloud.githubusercontent.com/assets/5943573/12871702/cd472d6e-cd3b-11e5-885e-e3cb9dd3aa7a.gif)

vasily-kirichenko:
I've never heard about *. PR is very much appreciated.

latkin:
Working on PR. Upstream FCS API needs 2 changes first
 - Report number of args associated with each specifier range
 - Start reporting `%%` as a specifier with 0 args

Will start there.  When considering `*`, one specifier can be associated with 0, 1, 2, or 3 args.

PR #1341, opened 2016-02-16
Handle 0-arity and multi-arity printf specifiers

latkin:
Fixes https://github.com/fsprojects/VisualFSharpPowerTools/issues/1335

specifiers


vasily-kirichenko:
Great.

PR #1349, opened 2016-03-17
Optimize on caret move

vasily-kirichenko:
Moving cursor with keyboard was a bit sluggish compared to C# editor, so I profiled and found that `dte.GetProjectItem` is the problem and add caching `IProjectProvider` in all places with `lazy`.

While I was at it, I tried to fix bug where symbol is not initially highlighted if you launch Go to definition. I found that `view.Caret.Position` here https://github.com/vasily-kirichenko/FSharpVSPowerTools/blob/optimize-on-caret-move/src/FSharpVSPowerTools.Logic/HighlightUsageTagger.fs#L97 is zero, so the feature does not work.

@dungpa Any ideas? It seems the caret moved event is not fired if the caret is moved with Go to definition.

### DIFF
--- a/src/FSharpVSPowerTools.Logic.VS2015/PeekDefinition.fs
+++ b/src/FSharpVSPowerTools.Logic.VS2015/PeekDefinition.fs
@@ -52,33 +52,24 @@ type PeekableItemSource
         buffer: ITextBuffer,
         doc: ITextDocument,
         peekResultFactory: IPeekResultFactory,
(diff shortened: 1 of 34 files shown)


dungpa:
Does the error happen on our own Go to definition or Visual F# tools' Go to definition? If it is on our own code, we could try to set caret position explicitly after going to definition?

vasily-kirichenko:
It's VS one :(

dungpa:
The changes look good.

PR #1364, opened 2016-04-15
Remove deprecated navigation bar setting

dungpa:
The setting has been deprecated from Visual F# tools.

Close https://github.com/fsprojects/VisualFSharpPowerTools/issues/1295.


vasily-kirichenko:
So no navigation bar anymore?..

smoothdeveloper:
@dungpa, I still see the navbar in my VS2015, what is up with removing it from settings? is it enabled by default in Visual F# tooling?

dungpa:
In Visual F# tools vNext, it will be either be available via Roslyn-based API or be removed completely. Either way, it's no longer Ok to check for the setting to enable the feature anymore.

The toggle for Navigation Bar in VFPT has never fully worked. I'm really happy when we can finally remove it.

jackfoxy:
Somehow the nav bar is mysteriously working in one of my 3 environment. All 3 environment are VS 2015 update 2 F#PT 2.4.0.

Go figure.

I personally would like this feature to remain as long as possible. Just because it is deprecated does not mean it does not still work. The fact it is still working in one of my environment is evidence of that. I'm not sure what the definition of "never fully worked" is. I do know it would appear/disappear with new PT releases sometimes. Still, it was worth the trouble for me. It's not like it cause VS to crash (at least I don't think it did).

dsyme:
The code from the Visual F# Tools is here: https://github.com/Microsoft/visualfsharp/blob/master/vsintegration/src/FSharp.LanguageService/NavigationBar.fs plus some C# code too in FSharp.LanguageService.Base

Maybe it could just be moved into VFPT.

PR #1365, opened 2016-04-16
Fixing an incorrect background color when non-standard theme is used with Depth Colorizer

### LINKED ISSUE #1358: VFPT overrides the default background color in code editor
### Description

When VFPT is enabled, it paints the background color in white, overriding the theme's color settings. This doesn't happen when VFPT is disabled.

I'm using a non-standard color scheme ([Jeff Atwood's Theme](https://blog.codinghorror.com/is-your-ide-hot-or-not/), which has the background color #F8F8F8.
When VFPT is enabled, it paints the background in white. In fact, the paining happens a second later after the code is change, so it's visible:
animation

### Repro steps

1. Install the Jeff Atwood theme from the link above (import .vssettings) or simply change the text editor's background color to #F8F8F8 (R:248 G:248 B:248)
2. Open a .fs file with VFPT enabled.

### Expected behavior

Code background should match the VS settings.

### Actual behavior

Code background is white.

hmemcpy:
This fixes #1358 

This PR makes several changes to the way theme settings are retrieved. Instead of querying the registry, this relies on the built in VS theming service (available since VS2012). Unfortunately, the public PIA types for this service were never shipped, so a bit of dynamic magic is needed to get the service.

In addition, a refactoring of the default color matrix array - now relies on the background color of the editor, instead of having hardcoded values. Also fixes hardcoded registry access.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -10,6 +10,7 @@ nuget FSharpLint.Core prerelease
 nuget VSSDK.IDE.Deploy.14
 nuget FSharp.Management
 nuget FsPickler
+nuget FSharp.Interop.Dynamic
 
 group Test
 source http://nuget.org/api/v2
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -10,7 +10,6 @@ nuget FSharpLint.Core prerelease
 nuget VSSDK.IDE.Deploy.14
 nuget FSharp.Management
 nuget FsPickler
-nuget FSharp.Interop.Dynamic
 
 group Test
 source http://nuget.org/api/v2
--- a/paket.lock
+++ b/paket.lock
@@ -1,15 +1,10 @@
 NUGET
   remote: http://www.nuget.org/api/v2
   specs:
-    Dynamitey (1.0.2)
     Expression.Blend.Sdk (1.0.2)
     Fantomas (2.1)
       FSharp.Compiler.Service (>= 2.0.0.8)
     FSharp.Compiler.Service (2.0.0.8)
-    FSharp.Core (4.0.0.1)
-    FSharp.Interop.Dynamic (3.0)
-      Dynamitey (>= 1.0.2)
-      FSharp.Core (>= 3.1.2.1)
     FSharp.Management (0.3)
     FSharp.ViewModule.Core (0.9.9.1)
     FSharpLint.Core (0.0.16)
--- a/paket.lock
+++ b/paket.lock
@@ -1,10 +1,15 @@
 NUGET
   remote: http://www.nuget.org/api/v2
   specs:
+    Dynamitey (1.0.2)
     Expression.Blend.Sdk (1.0.2)
     Fantomas (2.1)
       FSharp.Compiler.Service (>= 2.0.0.8)
     FSharp.Compiler.Service (2.0.0.8)
+    FSharp.Core (4.0.0.1)
+    FSharp.Interop.Dynamic (3.0)
+      Dynamitey (>= 1.0.2)
+      FSharp.Core (>= 3.1.2.1)
     FSharp.Management (0.3)
     FSharp.ViewModule.Core (0.9.9.1)
     FSharpLint.Core (0.0.16)
--- a/src/FSharpVSPowerTools.Logic/ThemeManager.fs
+++ b/src/FSharpVSPowerTools.Logic/ThemeManager.fs
@@ -6,12 +6,12 @@ open Microsoft.VisualStudio.Shell
 open Microsoft.VisualStudio.Shell.Interop
 open FSharpVSPowerTools
 open FSharpVSPowerTools.ProjectSystem
-open FSharp.Interop.Dynamic
 open System.Drawing
 open Microsoft.Win32
 open EnvDTE
 open Microsoft.VisualStudio.PlatformUI
 open System.Runtime.InteropServices
+open Reflection
 
 type VisualStudioTheme =
     | Unknown = 0
@@ -36,9 +36,7 @@ type ThemeManager [<ImportingConstructor>]
 
     member __.GetCurrentTheme() =
         let themeGuid = getThemeId()
-        match themes.TryGetValue(themeGuid) with
-        | true, t -> Some t
-        | _ -> None
+        themes |> Dict.tryFind themeGuid
         |> Option.getOrTry (fun _ ->
             try 
                 let color = VSColorTheme.GetThemedColor EnvironmentColors.ToolWindowTextColorKey
--- a/src/FSharpVSPowerTools.Logic/ThemeManager.fs
+++ b/src/FSharpVSPowerTools.Logic/ThemeManager.fs
@@ -6,16 +6,21 @@ open Microsoft.VisualStudio.Shell
 open Microsoft.VisualStudio.Shell.Interop
 open FSharpVSPowerTools
 open FSharpVSPowerTools.ProjectSystem
+open FSharp.Interop.Dynamic
 open System.Drawing
 open Microsoft.Win32
 open EnvDTE
 open Microsoft.VisualStudio.PlatformUI
+open System.Runtime.InteropServices
 
 type VisualStudioTheme =
     | Unknown = 0
     | Light = 1
     | Blue = 2
     | Dark = 3
+
+[<Guid("0d915b59-2ed7-472a-9de8-9161737ea1c5")>]
+type SVsColorThemeService = interface end
 
 [<Export>]
 type ThemeManager [<ImportingConstructor>] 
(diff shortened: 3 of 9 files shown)


paket.dependencies
```diff
@@ -10,6 +10,7 @@ nuget FSharpLint.Core prerelease
 nuget VSSDK.IDE.Deploy.14
 nuget FSharp.Management
 nuget FsPickler
+nuget FSharp.Interop.Dynamic
```
dungpa: Adding this dependency for a single use seems a bit overkill. We have used dynamic in the code base using this https://github.com/fsprojects/VisualFSharpPowerTools/blob/e3b2531d83e008360c1fd6bcdb0bcfb25c9aaa31/src/FSharpVSPowerTools.Logic/ProjectSystem.fs#L15. Could it eliminate the need of an extra dependency?

src/FSharpVSPowerTools.Logic/ThemeManager.fs
```diff
@@ -26,41 +31,14 @@ type ThemeManager [<ImportingConstructor>]
...
-            |> Option.bind (function 
-                            | [|_; _; themeId|] -> Some themeId 
-                            | arr -> 
-                                Logging.logWarning (fun _ -> sprintf "Parsed Visual Studio theme settings are not well-formed %A." arr)
-                                None)
-        | _ ->
-            Logging.logWarning (fun _ -> sprintf "Can't recognize Visual Studio version %A." version)
-            None
+        let themeService = serviceProvider.GetService(typeof<SVsColorThemeService>)
+        themeService?CurrentTheme?ThemeId: Guid
```
dungpa: How robust are these dynamic calls? Could they throw exceptions in certain cases? If yes, it would be good to wrap them under `Option.attempt` just to be sure.

src/FSharpVSPowerTools.Logic/ThemeManager.fs
```diff
@@ -26,41 +31,14 @@ type ThemeManager [<ImportingConstructor>]
...
-        |> Option.bind (fun themeId ->
-            match Guid.TryParse(themeId) with
-            | true, themeGuid ->
-                match themes.TryGetValue(themeGuid) with
-                | true, t -> Some t
-                | _ -> None
-            | _ -> None)
+        let themeGuid = getThemeId()
+        match themes.TryGetValue(themeGuid) with
+            | true, t -> Some t
```
dungpa: Style: `|` should align with `match` in the previous line (to be consistent with the rest of the code base).

src/FSharpVSPowerTools.Logic/ThemeManager.fs
```diff
@@ -71,3 +49,13 @@ type ThemeManager [<ImportingConstructor>]
             with _ -> 
                 Logging.logError (fun _ -> "Can't read Visual Studio themes from environment colors.")
                 VisualStudioTheme.Unknown)
+
+    member __.GetEditorTextColors (item: string) =
+        let dte = serviceProvider.GetDte()
+        let fontsAndColors = dte.Properties("FontsAndColors", "TextEditor")
+        let fontsAndColorsItems = fontsAndColors.Item("FontsAndColorsItems").Object :?> EnvDTE.FontsAndColorsItems
+        let selectedItem = fontsAndColorsItems.Item(item)
+        let foreColor =  int selectedItem.Foreground |> System.Drawing.ColorTranslator.FromOle
```
dungpa: Nits: remove an extra space before `int`.

paket.dependencies
hmemcpy: Possibly, I searched for how to do dynamic in F#, and `FSharp.Interop.Dynamic` was the first suggestion :)

Anyway, I'd be happy to replace it with plain old reflection, unfortunately I can't yet figure out how to do this in F#... Is there an example of how this function is used?

paket.lock
```diff
@@ -1,10 +1,15 @@
 NUGET
   remote: http://www.nuget.org/api/v2
   specs:
+    Dynamitey (1.0.2)
     Expression.Blend.Sdk (1.0.2)
     Fantomas (2.1)
       FSharp.Compiler.Service (>= 2.0.0.8)
     FSharp.Compiler.Service (2.0.0.8)
+    FSharp.Core (4.0.0.1)
```
hmemcpy: Same here, by the way, adding the `FSharp.Interop.Dynamic` dependency caused it to add FSharp.Core (4.0.0.1). If this is possible to remove, all the better.

@forki, do you have a better option?

src/FSharpVSPowerTools.Logic/ThemeManager.fs
hmemcpy: They would be as robust as reflection calls, I'd guess. But you're right, it's better to be safe. I'll try wrapping those in `Option`.

paket.dependencies
dungpa: It has been used extensively in this module https://github.com/fsprojects/VisualFSharpPowerTools/blob/e3b2531d83e008360c1fd6bcdb0bcfb25c9aaa31/src/FSharpVSPowerTools.Logic/FolderMenuCommands.fs.

paket.lock
forki: This is a bit worrying. @dungpa and @vasily-kirichenko what do you think

hmemcpy: @forki Regardless, I think I might just pull it off with the Reflection module @dungpa linked above, and no need for this dependency. I'll revert.

vasily-kirichenko: Yes, it's very worrying. However, I think it's time to switch to F# 4 and FSharp.Core anyway as it has some optimizations which are always very welcome in this project. It's a quite unrelated suggestion though.

src/FSharpVSPowerTools.Logic/ThemeManager.fs
```diff
@@ -26,41 +31,14 @@ type ThemeManager [<ImportingConstructor>]
...
-        getThemeId()
-        |> Option.bind (fun themeId ->
-            match Guid.TryParse(themeId) with
-            | true, themeGuid ->
-                match themes.TryGetValue(themeGuid) with
-                | true, t -> Some t
-                | _ -> None
-            | _ -> None)
+        let themeGuid = getThemeId()
+        match themes.TryGetValue(themeGuid) with
```
vasily-kirichenko: Use `Dict.tryFind` instead https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Core/Utils.fs#L746

forki: themes is IDictionary. that's why I created https://github.com/fsprojects/VisualFSharpPowerTools/pull/1367

hmemcpy: @forki @vasily-kirichenko yup, thanks! `themes` was an IDictionary, and it was incompatible with that Util. Thanks for fixing!

src/FSharpVSPowerTools.Logic/ThemeManager.fs
```diff
@@ -26,41 +31,15 @@ type ThemeManager [<ImportingConstructor>]
...
-                                Logging.logWarning (fun _ -> sprintf "Parsed Visual Studio theme settings are not well-formed %A." arr)
-                                None)
-        | _ ->
-            Logging.logWarning (fun _ -> sprintf "Can't recognize Visual Studio version %A." version)
-            None
+        match Option.attempt (fun _ -> 
+            let themeService = serviceProvider.GetService(typeof<SVsColorThemeService>)
+            themeService?CurrentTheme?ThemeId) with
+        | Some x -> x: Guid
+        | None -> Guid.Empty
```
hmemcpy: Ok, I tried using `Option.attempt`, but I really don't like this part, I'm sure can hide the match part behind some other Option helper. I want my function to return a Guid, not a Guid option.

hmemcpy:
Ok, I reverted using the Dynamic dependency, and moved the Reflection (?) operator to common utils. Please review :) @vasily-kirichenko @dungpa

src/FSharpVSPowerTools.Logic/ThemeManager.fs
vasily-kirichenko: https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Core/Utils.fs#L358

hmemcpy: Of course! I tried it with just orElse....

hmemcpy: Much better! Thank you @vasily-kirichenko!

dungpa:
The PR looks good.

@vasily-kirichenko Could you try this (theme change handling) out in VS 2015 before I merge the PR? Thanks.

vasily-kirichenko:
@dungpa sure. Let me ten minutes.

vasily-kirichenko:
1. Launch VS with the Blue theme.
2. Switch to the Dark one:

image

3. Switch to the Blue theme back:

image

dungpa:
@vasily-kirichenko Thanks. The bug on Functions category shows up again; it may be unrelated.

dungpa:
@hmemcpy It's a good idea to move all reflections functions into a single place. Thanks for the help.

PR #1367, opened 2016-04-18
use flexible types for dict module

forki:
/cc @hmemcpy


vasily-kirichenko:
Why?

forki:
I *believe* this will help @hmemcpy with his current pull request. (I should have sent this as PR to his fork)

PR #1374, opened 2016-04-25
fixed: UnusedDeclarations.getSingleDeclarations returns actually used…

vasily-kirichenko:
… symbol due to wrong equality function used for grouping `FSharpSymbolUse`s

This fixes a huge bug in unused declarations searching algorithm, which caused many used symbols to be marks as unused.

### DIFF
--- a/src/FSharpVSPowerTools.Core/TypedAstUtils.fs
+++ b/src/FSharpVSPowerTools.Core/TypedAstUtils.fs
@@ -300,27 +300,31 @@ module UnusedDeclarations =
     open System.Collections.Generic
 
     let symbolsComparer =
-        { new IEqualityComparer<FSharpSymbol> with
-              member __.Equals (x, y) = x.IsEffectivelySameAs y
-              member __.GetHashCode x = x.GetHashCode() }
+        { new IEqualityComparer<FSharpSymbolUse> with
+              member __.Equals (x, y) = x.Symbol.IsEffectivelySameAs y.Symbol
+              member __.GetHashCode x = x.Symbol.GetHashCode() }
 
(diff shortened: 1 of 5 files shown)


src/FSharpVSPowerTools.Core/TypedAstUtils.fs
```diff
@@ -300,27 +300,31 @@ module UnusedDeclarations =
     open System.Collections.Generic
 
     let symbolsComparer =
```
dungpa: Should it be `symbolUseComparer`?

src/FSharpVSPowerTools/FSharpVSPowerTools.csproj
```diff
@@ -46,7 +46,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <DeployExtension>True</DeployExtension>
+    <DeployExtension>False</DeployExtension>
```
dungpa: Why do you need to turn this off? It no longer allows deploying the extension, which is quite counter-intuitive.

vasily-kirichenko:
All done.

dungpa:
Great.

src/FSharpVSPowerTools/FSharpVSPowerTools.csproj
vasily-kirichenko: Two reasons:
* Having this on makes building in VS 2015 much slower. No idea why.
* Exp VS 2015 instance does not work anyway (I mean absolutely, never works), so it does not make any sense to deploy VSIX to it during each building.

PR #1384, opened 2016-05-10
do not cache IProjectProvider

vasily-kirichenko:


dungpa:
Good rollback.

I think caching IProjectProvider is more trouble than worth.

Since we have problems with slow GetProjectItem https://github.com/fsprojects/VisualFSharpPowerTools/blob/d3d386a6702cb761e3f40aa9b1ac4e4071c0c997/src/FSharpVSPowerTools.Logic/ProjectFactory.fs#L130, a better solution would be to use IVSRunningDocumentTable https://msdn.microsoft.com/en-us/library/bb166372.aspx

PR #1386, opened 2016-05-11
Remove FindProjectItem which is potentially slow

dungpa:
Related comment: https://github.com/fsprojects/VisualFSharpPowerTools/pull/1384#issuecomment-218116822

I don't have a big solution to try out the changes.

Could you try this PR on big solutions to see any performance enhancement compared to master?


vasily-kirichenko:
I've profiled this PR on my large solution and nothing related to `IVsRunningDocumentTable` popped up. Great job!

Issue #1370, opened 2016-04-22
Visual Studio 15 support


xperiandri:

vasily-kirichenko:
go ahead.

xperiandri:
Is there a sample project on which I can test all the features?
I just updated manifest and it works in VS 15 but I cannot be sure that all the features work.

vasily-kirichenko:
Why don't you test it on *any* project?

xperiandri:
I have 2. But they don't have all conditions for all features. I will need to simulate them.

jackfoxy:
After VS2015 update 2 I no longer see the navigation bar. I tried turning the option off and turning it back on, but that does not fix it.

xperiandri:
Works fine within Visual Studio 2015 Update 2 and in 15 Preview
Try devenv /setup

jackfoxy:
Nav bar has reappeared in one environment...
Ignore for now, must be something on my end.

calwi:
Same problem after Update 2 install, Navigation Bar had vanished from the options. I downloaded a new copy of the vsix to reinstall... removed the extension completely... now installer only gives option for VS2013 (which is also on the machine). So I don't have anything. Nor can I find it from within VS Tools & Extensions.

xperiandri:
Open it with 7-zip
Edit extension.vsixmanifest
```
    <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[13.0,16.0)" />
...
    <Dependency Id="FSharp.Editor" DisplayName="Microsoft Visual FSharp Editor Extensions" Version="[12.0,16.0)" Location="http://www.microsoft.com/en-us/download/details.aspx?id=48179" />
```
Close editor
Save
Install

calwi:
Sorry for not getting back sooner. @xperiandri solution hasn't worked. I'll try this from scratch on a fresh vm and install and see if anything alters.

xperiandri:
Pick Definition is grayed out.
Displays 2015 only

xperiandri:
```
[...]

PR #1389, opened 2016-05-12
Add support for Visual Studio 15

jakubsuchybio:
Regarding #1370, 
I couldn't find your extension in VS Gallery for VS15Preview. 
So this PR is just updating vsmanifest to take the new VS15Preview into consideration.
I installed it to my VS15 and everything seems fine. 
However I was not able to run your tests, because they couldn't be found via Test Explorer.


vasily-kirichenko:
About the tests, you need to install NUnit tests adapter addin to see them in Test Explorer.

vasily-kirichenko:
Thanks!

PR #1401, opened 2016-07-09
Refactor HighlightUsageTagger

smoothdeveloper:
So that logic is separated from FSharpVSPowerTools.Logic project with a new HighlightUsageInFile module in FSharpVSPowerTools.Core project

### DIFF
--- a/src/FSharpVSPowerTools.Core/Symbols/HighlightUsageInFile.fs
+++ b/src/FSharpVSPowerTools.Core/Symbols/HighlightUsageInFile.fs
@@ -17,9 +17,9 @@ module Symbols =
 
 type FileName = string
 type GetCheckResults = FileName -> Async<ParseAndCheckResults option>
-type FCSZeroBasedRange = int * int * int * int
+type ZeroBasedRange = int * int * int * int
 
-type CurrentLine = {Line: string; File: FileName; Range: FCSZeroBasedRange}
+type CurrentLine = { Line: string; File: FileName; Range: ZeroBasedRange }
     with
         member x.EndLine =
             let _, _, endLine, _ = x.Range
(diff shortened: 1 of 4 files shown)


src/FSharpVSPowerTools.Core/Symbols/HighlightUsageInFile.fs
```diff
@@ -0,0 +1,41 @@
...
+        |> Seq.groupBy (fst >> Path.GetFullPathSafe)
+        |> Seq.collect (fun (_, symbolUses) -> 
+            symbolUses 
+            |> Seq.map snd 
+            |> Seq.distinctBy (fun s -> s.RangeAlternate))
+        |> Seq.toArray
+
+type FileName = string
+type GetCheckResults = FileName -> Async<ParseAndCheckResults option>
+type FCSZeroBasedRange = int * int * int * int
```
dungpa: Remove `FCS` prefix. It doesn't add more information.

src/FSharpVSPowerTools.Core/Symbols/HighlightUsageInFile.fs
```diff
@@ -0,0 +1,41 @@
...
+
+type CurrentLine = {Line: string; File: FileName; Range: FCSZeroBasedRange}
+    with
```
dungpa: The use of `with` is weird.
You can use indentation:

```fsharp
type CurrentLine = 
    {Line: string; File: FileName; Range: FCSZeroBasedRange}
    member x.EndLine = ...
```

src/FSharpVSPowerTools.Core/Symbols/HighlightUsageInFile.fs
```diff
@@ -0,0 +1,41 @@
...
+        member x.EndLine =
+            let _, _, endLine, _ = x.Range
+            endLine
+
+module HighlightUsageInFile =
+    open FSharpVSPowerTools
+    open Microsoft.FSharp.Compiler.SourceCodeServices
+    [<NoComparisonAttribute(* due to FSharpSymbol*) >]
+    type HighlightUsageInFileResult =
+    | UsageInFile of FSharpSymbol * string * FSharpSymbolUse array
```
dungpa: We often use indentation for union cases in the code base.

dungpa:
Looks good. Thanks for doing it.

src/FSharpVSPowerTools.Core/Symbols/HighlightUsageInFile.fs
smoothdeveloper: This doesn't seem to work with records, I assume those are seen as (intrinsic) extension members.

smoothdeveloper:
@dungpa thanks for review, I can't fix the record which requires `with`, please let me know if we should do it another way.

dungpa:
I have an example at https://github.com/dungpa/fantomas/blob/dbb019431910a8fbc416d5aa003d79ca0dd6960c/src/Fantomas/FormatConfig.fs#L118-L156.

But it's not important anyway.

PR #1402, opened 2016-07-09
GoToDefinition refactors first pass

smoothdeveloper:
I want to change a bit the logic split in getDocumentState and where it used (matching over large tuple, and more logic if not matching).


vasily-kirichenko:
May we merge my PR first since it restructuring Logic a lot?

smoothdeveloper:
sure, and I need to rebase my own master. This can go in then.

PR #1404, opened 2016-07-09
[wip] Refactor GoToDefinitionFilter 

smoothdeveloper:
getDocumentState/gotoDefinition with more appropriate separation of logic to identify definition type and how to handle it

Tried it on paket solution, go to script, go to symbol in same file and go to symbol in FSharp.Core all worked.

### DIFF
--- a/src/FSharpVSPowerTools.Core/Lexer.fs
+++ b/src/FSharpVSPowerTools.Core/Lexer.fs
@@ -89,7 +89,7 @@ module Lexer =
         loop (queryLexState source defines line) []
 
     // Returns symbol at a given position.
-    let getSymbolFromTokens (tokens: FSharpTokenInfo list) line col (lineStr: string) lookupKind: Symbol option =
+    let getSymbolFromTokens (tokens: FSharpTokenInfo list) line col (lineStr: string) lookupKind : Symbol option =
         let isIdentifier t = t.CharClass = FSharpTokenCharKind.Identifier
         let isOperator t = t.ColorClass = FSharpTokenColorKind.Operator
     
@@ -198,6 +198,7 @@ module Lexer =
                   RightColumn = token.RightColumn + 1
                   Text = lineStr.Substring(token.Token.LeftColumn, token.Token.FullMatchedLength) })
     
+    
     let getSymbol source line col lineStr lookupKind (args: string[]) queryLexState =
         let tokens = tokenizeLine source args line lineStr queryLexState
         try
@@ -206,6 +207,5 @@ module Lexer =
             debug "Getting lex symbols failed with %O" e
             None
 
-    let getSymbolAtPoint point lookupKind (args) queryLexState =
-        let line, col = point.Point
-        getSymbol point.Document line col point.Line lookupKind args queryLexState
+    let getSymbolAtPoint point lookupKind args queryLexState =
+        getSymbol point.Document point.Point.Line point.Point.Column point.Line lookupKind args queryLexState
--- a/src/FSharpVSPowerTools.Core/Lexer.fs
+++ b/src/FSharpVSPowerTools.Core/Lexer.fs
@@ -16,7 +16,7 @@ type Symbol =
       LeftColumn: int
       RightColumn: int
       Text: string }
-    member x.Range : Range<FCS> = { From = Point.make x.Line x.LeftColumn; To = Point.make x.Line x.RightColumn }
+    member x.Range : Range<FCS> = { Start = Point.make x.Line x.LeftColumn; End = Point.make x.Line x.RightColumn }
 
 [<RequireQualifiedAccess>]
 type SymbolLookupKind =
--- a/src/FSharpVSPowerTools.Core/Lexer.fs
+++ b/src/FSharpVSPowerTools.Core/Lexer.fs
@@ -1,5 +1,5 @@
 ﻿namespace FSharpVSPowerTools
-
+open FSharpPowerTools.Core
 open System.Diagnostics
 open Microsoft.FSharp.Compiler.SourceCodeServices
 
@@ -30,6 +30,11 @@ type internal DraftToken =
       RightColumn: int }
     static member inline Create kind token = 
         { Kind = kind; Token = token; RightColumn = token.LeftColumn + token.FullMatchedLength - 1 }
+
+type SourceCode = string
+type Defines = string list
+type LineIndex = int
+type QueryLexState = SourceCode -> Defines -> LineIndex -> FSharpTokenizerLexState
 
 module Lexer =
     /// Get the array of all lex states in current source
@@ -71,7 +76,7 @@ module Lexer =
             lexStates.[line]
 
     /// Return all tokens of current line
-    let tokenizeLine source (args: string[]) line lineStr queryLexState =
+    let tokenizeLine source (args: string[]) line lineStr (queryLexState: QueryLexState) =
         let defines =
             args |> Seq.choose (fun s -> if s.StartsWith "--define:" then Some s.[9..] else None)
                  |> Seq.toList
@@ -199,4 +204,8 @@ module Lexer =
             getSymbolFromTokens tokens line col lineStr lookupKind
         with e ->
             debug "Getting lex symbols failed with %O" e
-            None 
+            None
+
+    let getSymbolAtPoint point lookupKind (args) queryLexState =
+        let line, col = point.Point
+        getSymbol point.Document line col point.Line lookupKind args queryLexState
--- a/src/FSharpVSPowerTools.Core/Lexer.fs
+++ b/src/FSharpVSPowerTools.Core/Lexer.fs
@@ -16,7 +16,7 @@ type Symbol =
       LeftColumn: int
       RightColumn: int
       Text: string }
-    member x.Range = x.Line, x.LeftColumn, x.Line, x.RightColumn
+    member x.Range : Range<FCS> = { From = Point.make x.Line x.LeftColumn; To = Point.make x.Line x.RightColumn }
 
 [<RequireQualifiedAccess>]
 type SymbolLookupKind =
--- a/src/FSharpVSPowerTools.Core/Utils.fs
+++ b/src/FSharpVSPowerTools.Core/Utils.fs
@@ -5,7 +5,7 @@ type FileName = string
 [<Measure>] type FCS
 
 type Point<[<Measure>]'t> = { Line : int; Column : int }
-type Range<[<Measure>]'t> = { From : Point<'t>; To: Point<'t> }
+type Range<[<Measure>]'t> = { Start : Point<'t>; End: Point<'t> }
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module Point =
@@ -16,13 +16,13 @@ module Range =
 
     let make startLine startColumn endLine endColumn : Range<'t> =
         {
-          From = Point.make startLine startColumn
-          To = Point.make endLine endColumn 
+          Start = Point.make startLine startColumn
+          End = Point.make endLine endColumn 
         }
     
 type CurrentLine<[<Measure>]'t> = { Line: string; File: FileName; Range: Range<'t> }
     with
-        member x.EndLine = x.Range.To.Line 
+        member x.EndLine = x.Range.End.Line 
 
 [<NoComparison>]
 type PointInDocument<[<Measure>]'t> = {
--- a/src/FSharpVSPowerTools.Core/Utils.fs
+++ b/src/FSharpVSPowerTools.Core/Utils.fs
@@ -9,8 +9,7 @@ type Range<[<Measure>]'t> = { From : Point<'t>; To: Point<'t> }
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module Point =
-    let make    line column : Point<'t> = { Line = line; Column = column }
-    //let makeFCS line column : Point<FCS> = { Line = line; Column = column }
+    let make line column : Point<'t> = { Line = line; Column = column }
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module Range =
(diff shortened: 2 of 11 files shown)


src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
```diff
@@ -33,6 +28,12 @@ type internal UrlChangeEventArgs(url: string) =
     inherit EventArgs()
     member __.Url = url
 
+[<NoComparisonAttribute; NoEquality>]
```
dungpa: We could remove `Attribute` suffix here.

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
```diff
@@ -33,6 +28,12 @@ type internal UrlChangeEventArgs(url: string) =
...
+type GoToDefinitionResolution =
```
dungpa: I think a more fine-grained DU (adding `NotFound` case) could simplify the code by removing option values at their usage.

```fsharp
type GoToDefinitionResult =
    | Internal
    | External of IProjectProvider * ParsedInput option * SnapshotSpan * FSharpSymbolUse
    | LoadDirective of string
    | NotFound
```

I rename other union cases to denote where they come from.

dungpa:
This is a good idea. I had a few comments above.

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
smoothdeveloper: I was considering the NotFound case but in current implementation (in asyncMaybe) it still makes sense to return Some/None, I'll give it another look.

Also `SnapshotSpan` will be going away from `External` case as I'll be working on making the logic work without Visual Studio APIs.

smoothdeveloper: Also `LoadDirective` clashes (potentially) with other places having this identifier (it is a DU in Untyped AST utils).

dungpa: For big project, it makes sense to use qualified names e.g. `GoToDefinitionResult. LoadDirective`. You can add `RequireQualifiedAccess` to DU definition to reinforce that.

smoothdeveloper: I'm considering `FoundInternal`, `FoundExternal`, `FoundLoadDirective` and `NotFound`, what do you think?

Those will migrate to Core ultimately and will be consumed by client code, while LoadDirective normally won't.

dungpa: Looks good.

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
```diff
@@ -54,23 +55,30 @@ type GoToDefinitionFilter
...
+        asyncMaybe {
+            let fileName = doc.FilePath
+            let! project = project ()
+            let! caretPos = textBuffer.GetSnapshotPoint view.Caret.Position
+            let spanAndSymbol = vsLanguageService.GetSymbol(caretPos, fileName, project)
+            
+            match spanAndSymbol with
+            | Some (span, symbol) ->
+              let! result = vsLanguageService.GetFSharpSymbolUse(span, symbol, fileName, project, AllowStaleResults.MatchingSource) |> liftAsync
+              match result with
```
vasily-kirichenko: You can simplify it to 
```fsharp
let! fsSymbolUse, fileScopedCheckResults = 
    vsLanguageService.GetFSharpSymbolUse(span, symbol, fileName, project, AllowStaleResults.MatchingSource)
```

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
```diff
@@ -54,23 +55,30 @@ type GoToDefinitionFilter
...
+              | Some (fsSymbolUse, fileScopedCheckResults) ->
+                let lineText = span.Start.GetContainingLine().GetText()
+                let! findDeclResult = fileScopedCheckResults.GetDeclarationLocation(symbol, lineText, preferSignature = false) |> liftAsync
+              
+                let gotoDefinitionResult = 
+                    match findDeclResult with
+                    | FSharpFindDeclResult.DeclFound _ -> FoundByFSharpLanguageService
+                    | FSharpFindDeclResult.DeclNotFound _ -> FoundButNotHandled (project, fileScopedCheckResults.ParseTree, span, fsSymbolUse)
+              
+                return! Some gotoDefinitionResult
```
vasily-kirichenko: You can simplify it to:
```fsharp
return
    match findDeclResult with
    | FSharpFindDeclResult.DeclFound _ -> FoundByFSharpLanguageService
    | FSharpFindDeclResult.DeclNotFound _ -> FoundButNotHandled (project, fileScopedCheckResults.ParseTree, span, fsSymbolUse)
```

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
```diff
@@ -54,23 +55,30 @@ type GoToDefinitionFilter
...
+              | None -> return! None
+            | None -> 
+                let! loadDirectiveFile  = vsLanguageService.GetLoadDirectiveFileNameAtCursor(fileName, view, project)
+                return! Some (FoundLoadDirective loadDirectiveFile)
```
vasily-kirichenko:
```fsharp
return FoundLoadDirective loadDirectiveFile
```

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
smoothdeveloper: Somehow I'm getting this:
```
This expression was expected to have type
    (FSharpSymbolUse * ParseAndCheckResults) option    
but here has type
    'a * 'b
```

vasily-kirichenko: Look at my snippet - there is no liftAsync

src/FSharpVSPowerTools.Core/Lexer.fs
```diff
@@ -199,4 +204,8 @@ module Lexer =
             getSymbolFromTokens tokens line col lineStr lookupKind
         with e ->
             debug "Getting lex symbols failed with %O" e
-            None 
\ No newline at end of file
+            None
+
+    let getSymbolAtPoint point lookupKind (args) queryLexState =
```
vasily-kirichenko: Why parens here?

src/FSharpVSPowerTools.Core/Lexer.fs
```diff
@@ -199,4 +204,8 @@ module Lexer =
...
+        let line, col = point.Point
```
vasily-kirichenko: What about proper type for points instead of tuples?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -1,4 +1,28 @@
...
+type PointInDocument = {
+  Point: ZeroBasedPoint
+  Line: string
+  Document: string
+  File: FileName
+}
+    with
+        member x.LineIndex = fst x.Point
+        member x.ColumnIndex = snd x.Point
+        member x.CurrentLine = {Line = x.Line; File = x.File; Range = x.LineIndex, x.ColumnIndex, x.LineIndex, x.Line.Length }
```
vasily-kirichenko: I don't like you allocate a record each time the property is accessed. Also, I think it'd be simplified if you introduce `type Point = { Line: int; Col: int }` and use it in all these types.

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
```diff
@@ -184,26 +187,23 @@ type GoToDefinitionFilter
...
-                | Some (project, parseTree, span, fsSymbolUse, FSharpFindDeclResult.DeclNotFound _) ->
+                    // Run the operation on UI thread since continueCommandChain may access UI components.
+                    do! Async.SwitchToContext uiContext
+                    // Declaration location might exist so let Visual F# Tools handle it. 
+                    return continueCommandChain()
+                | Some (FoundLoadDirective fileToOpen) -> 
+                    // directive found, navigate to the file at first line
+                    serviceProvider.NavigateTo(fileToOpen, startRow=0, startCol=0, endRow=0, endCol=0)
+                | Some (FoundExternal (project, parseTree, symbol, fsSymbolUse)) ->
+                    let snapshotPoint = snapshotPoint |> Option.get // this Option.get is guaranted safe because we actually had a match at position
```
vasily-kirichenko: Sorry, it's not clear that `snapshotPoint` is `Some` here.

src/FSharpVSPowerTools.Logic/ProjectSystem/VSLanguageService.fs
```diff
@@ -338,4 +348,17 @@ type VSLanguageService
     /// This value is used for testing when VS lex cache isn't available
     member internal __.SkipLexCache 
         with get () = skipLexCache
-        and set v = skipLexCache <- v
\ No newline at end of file
+        and set v = skipLexCache <- v
+
+
+    member __.MakePointInDocument (point: SnapshotPoint option) filename =
```
vasily-kirichenko: I don't think this method should belong to `VSLanguageService`.

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
smoothdeveloper: It is because before we call the logic in getDocumentState, it would exit with None if snapshotPoint was `None`. What do you recommend?

src/FSharpVSPowerTools.Core/Utils.fs
smoothdeveloper: Is Lazy a good solution for CurrentLine property?

I can make a proper point type.

src/FSharpVSPowerTools.Logic/ProjectSystem/VSLanguageService.fs
smoothdeveloper: I'm thinking as extension member to SnapshotPoint (which is object we will use in most cases to construct PointInDocument)?

src/FSharpVSPowerTools.Core/Utils.fs
vasily-kirichenko: I think proper `Point` type should be used thought the codebase. Passing `line` and `col` everywhere, often as curred arguments is strange, unnecessary error prone and verbose.

src/FSharpVSPowerTools.Logic/Navigation/GoToDefinitionFilter.fs
vasily-kirichenko: Logically you are right, the code is correct, but the correctness is not enforced by the compiler. It's not good.

smoothdeveloper: Understand, it felt weird as well and I was not trying to conceal that logical thing, would you be ok with:

```fsharp
match snapshotPoint with
| None -> return ()
| Some snapshotPoint ->
   // rest of the code indented
```

vasily-kirichenko: Yes, it's much better.

src/FSharpVSPowerTools.Core/Utils.fs
smoothdeveloper: This means large refactoring, and we should also clarify onebased / zerobased stuff which is error prone and make contributors spend time figuring when/where to do the conversion.

I think it makes more sense to make most of our code to be "close to FCS" and keep VS stuff only "close to VS" and convert to structures which are VS agnostic (we shouldn't cary SnapshotPoint etc. through several calls/layers).

Can we deal with this later though (I'll make a `Point` type and also use Lazy for that property)? I'd like to complete refactoring the logic for Go To Definition to move it to Core and focus on maybe other features.

vasily-kirichenko: I disagree that we should be close to FCS. FCS has an ugly API and we should hide it as much as possible. What's more, Don tends to change the API, which sometimes breaks our codebase in many places. 

Yes, I think we can leave the Point type for later refactoring.

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -1,4 +1,42 @@
...
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Range =
+
+    let make startLine startColumn endLine endColumn : Range<'t> =
+        {
+          From = Point.make startLine startColumn
+          To = Point.make endLine endColumn 
+        }
+    
+type CurrentLine<[<Measure>]'t> = { Line: string; File: FileName; Range: Range<'t> }
```
vasily-kirichenko: Maybe `RangeInFileWithSourceLine`?

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -1,4 +1,42 @@
...
+    with
+        member x.EndLine = x.Range.To.Line 
+
+[<NoComparison>]
+type PointInDocument<[<Measure>]'t> = {
+  Point    : Point<'t>
+  Line     : string
+  Document : string
+  File     : FileName
+}
```
vasily-kirichenko: This type is very similar to `CurrentLine`. What about embed the latter to it? Like:
```fsharp
type PointInDocument =
    { Line: CurrentLine
      Document: string }
```

src/FSharpVSPowerTools.Core/Utils.fs
```diff
@@ -1,4 +1,42 @@
-﻿namespace FSharpVSPowerTools
+﻿namespace FSharpPowerTools.Core
+
+type FileName = string
+
+[<Measure>] type FCS
+
+type Point<[<Measure>]'t> = { Line : int; Column : int }
+type Range<[<Measure>]'t> = { From : Point<'t>; To: Point<'t> }
```
vasily-kirichenko: I cannot find any usages of these types specified with other than `FCS` measure. What's the point in measures here?

vasily-kirichenko: In ranges `Start` and `End` are usually used instead of `From` and `To`.

Issue #1411, opened 2016-07-13
ProjectProvider & ProjectFactory -> VFPT.Core


cloudRoutine:
_Dependent upon completion of Open Documents Tracker_ #1410 

`IProjectProvider` interface will be defined and the `ProjectProvider` type will be implemented in VFPT.Core. (Logic may still need to have a `VsProjectProvider` that implements the `IProjectProvider` interface). The `ProjectProvider` implementation can probably be changed to acquire compilation flags, project contents and configurations settings, 

_WIP - Need to review further to layout how to split roles between `ProjectFactory` in Core and Logic_

Code Located @ 
https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/ProjectSystem/ProjectProvider.fs
https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/ProjectSystem/ProjectFactory.fs

vasily-kirichenko:
I'd like to rename it to `ISolution`/`Solution`.

smoothdeveloper:
@cloudRoutine having this moved to core would simplify work I'm doing on GoToDefinitionFilter as i'd need not to get rid of it / `ISolution`.

So for now, I won't move stuff to Core but focus on other remaining dependencies.

smoothdeveloper:
@vasily-kirichenko I agree we should have something for `Solution` but we can't just rename it I believe, for example `GetReferencedProjects` doesn't make sense on `Solution` and probably other inconsistencies.

I think we can start define `Solution` and move the "identified once for all projects in solution" aspects to that interface when they become identified.

What do you all think?

vasily-kirichenko:
Solution concept for us is just "set of projects", nothing more. I don't think it should do anything beyond simple operations like "opened", "closed", "project loaded", "project updated", etc.

smoothdeveloper:
We can split out ProjectDescriptor (a record with the same member as the get properties in current IProjectProvider) as I was (kind of) describing in #1440, I think this should be done disregarding the fact that #1410 is done or not.

PR #1416, opened 2016-07-14
move IProjectProvider interface to Core project

smoothdeveloper:
this is a step for #1411 and is also removing one thing to take care in #1404

### DIFF
--- a/src/FSharpVSPowerTools.Logic/ProjectSystem/VSLanguageService.fs
+++ b/src/FSharpVSPowerTools.Logic/ProjectSystem/VSLanguageService.fs
@@ -19,26 +19,17 @@ open FSharpVSPowerTools.AssemblyContentProvider
 open FSharpVSPowerTools.AsyncMaybe
 open FSharpPowerTools.Core.Symbols
 open FSharpPowerTools.Core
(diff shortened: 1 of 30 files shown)


src/FSharpVSPowerTools.Logic/ProjectSystem/VSLanguageService.fs
```diff
@@ -19,27 +19,18 @@ open FSharpVSPowerTools.AssemblyContentProvider
 open FSharpVSPowerTools.AsyncMaybe
 open FSharpPowerTools.Core.Symbols
 open FSharpPowerTools.Core
+open FSharpPowerTools.Core.Infrastructure
 
 type FilePath = string
 
+type IProjectProvider = FSharpPowerTools.Core.Infrastructure.IProjectProvider
```
vasily-kirichenko: Could we just use the moved interface directly, without this alias?

smoothdeveloper: I now removed the type alias and added the open namespace

vasily-kirichenko:
Thanks, great :+1:

Issue #1410, opened 2016-07-13
Open Documents Tracker -> VFPT.Core


cloudRoutine:
In order to move the functionality of some of the features over to VFPT.Core we need an open documents tracker on the Core side that will store the dirty buffers of the documents currently open in the editor. This will not remove the `OpenDocumentsTracker` from the Logic side, but will shift it's purpose to sending Document Open, Closed, and Changed events as well as the text of these documents to the Core side.  It may need to apply changes to open docs received back from the Core side, but it's  currently unclear if that will be necessary.

In VFPT.Core the `OpenDocument` record will store the text and the range equivalent to the snapshot span instead of `ITextDocument` and `ITextSnapshot`

The types in logic should be renamed to `VSOpenDocument` and `VSOpenDocumentsTracker`

Code Located @ 
https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/src/FSharpVSPowerTools.Logic/ProjectSystem/OpenDocumentsTracker.fs

PR #1418, opened 2016-07-14
Move iopendocumentstracker

smoothdeveloper:
First step for #1410


src/FSharpVSPowerTools.Logic/ProjectSystem/OpenDocumentsTracker.fs
```diff
@@ -15,23 +16,24 @@ type OpenDocument =
...
       LastChangeTime: DateTime
       ViewCount: int }
-    member x.Text = lazy (x.Snapshot.GetText())
     static member Create document snapshot encoding lastChangeTime = 
         { Document = document
           Snapshot = snapshot
           Encoding = encoding
           LastChangeTime = lastChangeTime
           ViewCount = 1 }
+    member private x.text = lazy (x.Snapshot.GetText())
```
vasily-kirichenko: Why this extra property? Why not just use x.Text?

smoothdeveloper: Can't use Text without changing call sites due to how interfaces are explicitly implemented in F#, I'd need to change call sites and lose access to other properties.

vasily-kirichenko:
It failed to compiled.

smoothdeveloper:
@vasily-kirichenko sorry I forgot the added file, it seems to pass now.

vasily-kirichenko:
Great.

Issue #1425, opened 2016-07-15
Xml Documentation of VS SDK


smoothdeveloper:
We have VS SDK assemblies in those folders : 

https://github.com/fsprojects/VisualFSharpPowerTools/tree/master/lib/vs2015

it would be convenient to add the .xml so we have documentation while editing the code.

dungpa:
My suggestion is to move all these assemblies to NuGet packages and handle them via Paket https://www.nuget.org/profiles/VisualStudioExtensibility.

smoothdeveloper:
Ok, I'll do a PR for this.

smoothdeveloper:
I've been as far as having all projects to compile off the nuget packages, with all the separate version of VSSDK, although couldn't figure out the `Microsoft.TC.*` ones so those remain in the lib folder (as well as the native files).

Now the VSIX build still gives an error which I'm not able to figure out at this time.

dungpa:
The test assemblies (Microsoft.TC.*) are no longer used. They can be removed I think.
Native assemblies might come with SourceLink NuGet package though I'm not sure. They can stay there if we can't find them anywhere.

smoothdeveloper:
@dungpa TestUtilities project (which I assume is not really our code) doesn't compile without those. Our own VS test assembly depends on that project.

dungpa:
You can choose to remove the classes that use these references if it's easy; otherwise, just leave them intact in lib folder.  We don't use this piece of test functionalities at all.

PR #1437, opened 2016-07-16
VSSDK in paket.dependencies

smoothdeveloper:
attempt (almost complete!) to solve #1425

* make projects to compile against VS SDK nuget packages
* cleanup duplicate references in different groups (such as FAKE)
* reorganize all paket files contents

the vsix doesn't package itself due to weird errors that needs investigation


vasily-kirichenko:
Rebase it on master.

smoothdeveloper:
Ok, now all the "normal" compile time build failures are fixed on appveyor, only the vsix packaging is failing (same as what I'm seeing locally).

I wonder if someone with a bit more of experience with VSSDK could give it a quick look (don't spend hours) to see if this can be solved.

cc @jaredpar any idea how to diagnose this `Given type must derive from ITag`:

https://ci.appveyor.com/project/dungpa/visualfsharppowertools/build/2.6.0-a1607161349#L735

dungpa:
As the first step, you could:
 - merge `vssdklegacy` and `vssdk10` groups.
 - remove `vssdk11` and `vssdk12` groups completely by changing references to vssdk14.

This will simplify package management of VSSDK assemblies.

smoothdeveloper:
@dungpa are we dropping support for VS2013?

dungpa:
Although we compile for VS2015, binaries can still be used for VS2013.

smoothdeveloper:
Ok, I'm giving it another shot, I'm thinking, do we need a separate v2015 project if the SDK/VS doesn't have issue with code calling later part of SDK?

Say we develop feature for next release VS15, does that have to be in separate project?

smoothdeveloper:
Problem if I remove vssdk12 (and keep only vssdklegacy and a merged vssdk) is that I get obsolete feature warning (as error) at each usage of `SmartTag`.

I don't know how to suppress those wanrnings.

smoothdeveloper:
@dungpa @vasily-kirichenko this is green :)

I had to add 44 at the warning exclusion list though (Obsolete members), I hope we are ok with this.

paket.dependencies
```diff
@@ -1,35 +1,59 @@
...
+    nuget VSSDK.NavigateTo
+    nuget VSSDK.VSLangProj
+    
+    
+group vssdk
+    source http://nuget.org/api/v2
+    framework >= net45
+    
+    nuget Microsoft.VisualStudio.Editor
+    nuget Microsoft.VisualStudio.Language.Intellisense
```
jaredpar: Sorry not as familiar with Paket.  But one item here that stands out is the lack of version.  It's imperative that Visual Studio editor dependencies include version information.  The lack of versions + VS SDK installation + MSBuild resolution rules can easily lead to a project depending on a version of the VS editor binaries that it did not intend to.  Run into this many times myself. 

Can these be tied to a specific version of the binaries?  In particular i believe the editor ones should be 14.0.  


Due to the VS SDK installation + MSBuild resolution rules

src/FSharp.Editing.VisualStudio.v2015/FSharp.Editing.VisualStudio.v2015.fsproj
```diff
@@ -255,4 +211,212 @@
       </ItemGroup>
     </When>
   </Choose>
+  <Choose>
+    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersion) == 'v4.5.1' Or $(TargetFrameworkVersion) == 'v4.5.2' Or $(TargetFrameworkVersion) == 'v4.5.3' Or $(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3')">
+      <ItemGroup>
+        <Reference Include="Microsoft.VisualStudio.CoreUtility">
```
jaredpar: This is another case where it's important to have the full assembly name including version for VS editor dependencies.

paket.dependencies
smoothdeveloper: @jaredpar thanks for feedback!

You can see all versions in https://github.com/fsprojects/VisualFSharpPowerTools/blob/master/paket.lock

I think with the qualified paths (not based on assembly name resolution but actual path to the file) to the assemblies in *proj files it should pick up the correct versions right?

We actually simplified with input of @dungpa by having only vssdk14 (+ legacy for EnvDTE etc.) but I'm not sure yet it allows to run our extension in VS2013.

src/FSharp.Editing.VisualStudio.v2015/FSharp.Editing.VisualStudio.v2015.fsproj
smoothdeveloper: with the hintpath (and the file being present obviously) is there any chance msbuild would go somewhere else?

With the paket.lock file we have certitude of each package version, if we'd need several versions of one given nuget we'd put those in a different group (in paket parlance those are named set of direct dependencies) and they'll endup in different directories.

Here the group is `vssdk` and the file in hintpath is from this nuget package: `Microsoft.VisualStudio.CoreUtility (14.3.25407)`

paket.dependencies
jaredpar: @smoothdeveloper 

Again sorry about my Paket ignorance here.  But doesn't the lock file represent the result of a restore vs. specifying what packages are needed?  If so it seems that if a new version of these packages are uploaded then a restore could reference them and only change the lock file.  Is that correct?  

Basically I'm interested in what mechanism would stop the packages from silently upgrading to 15.0.  

> I think with the qualified paths (not based on assembly name resolution but actual path to the file) to the assemblies in *proj files it should pick up the correct versions right?

I'm not sure which value will take precedence in MSBuild: `HintPath` or values returned from the SDK.  I believe `HintPath` does take precedence today but I wouldn't bet on it, definitely wouldn't bet on it being true forever.

This is part of the reason I eventually switched to always using strongly named references for the VS editor binaries.  It allowed me to take MSBuild resolution out of the equation.  No matter what the rules are it will only succeed if it matches the exact assembly that I want :smile:

smoothdeveloper: > Again sorry about my Paket ignorance here. But doesn't the lock file represent the result of a restore vs. specifying what packages are needed?

no changes to paket.lock would only happen on `paket upgrade` or `paket install`, normal workflow (for someone building) invokes `paket restore` which will strictly abide to the `paket.lock` file (or fail if packages aren't available).

> Basically I'm interested in what mechanism would stop the packages from silently upgrading to 15.0.

the truth is in `paket.lock` and this file isn't touched unless someone invokes the `install` or `upgrade` commands, we can also specify `< 15` in `paket.dependencies` as well which will in turn avoid `install` or `upgrade` to look at eventual versions 15 or above.

> or values returned from the SDK.

Is it related to GAC? Is there any way to check that value?

>  I believe HintPath does take precedence today but I wouldn't bet on it, definitely wouldn't bet on it being true forever.

Is there a way to specify file reference in a more deterministic fashion? (fails if file not at location, no fallback).

> This is part of the reason I eventually switched to always using strongly named references for the VS editor binaries. It allowed me to take MSBuild resolution out of the equation. No matter what the rules are it will only succeed if it matches the exact assembly that I want :smile:

This is interesting, although it adds a lot of noise to the *proj files, also I think paket doesn't ever write the version in *proj files, I assume main reasons being:

* all versions are listed in sole `paket.lock` no need for redundant (and possibly mismatching) information
* code churn on project files
* assumption that assembly file reference would be determining which assembly ends up being referenced

jaredpar: > Is it related to GAC? Is there any way to check that value?

I don't believe so.  

> the truth is in paket.lock and this file isn't touched unless someone invokes the install or upgrade commands, we can also specify < 15 in paket.dependencies as well which will in turn avoid install or upgrade to look at eventual versions 15 or above.

I wish there was a stronger way to prevent this.  But if not then we just need to be very diligent about the contents of the lock file here.

smoothdeveloper: Thanks @jaredpar

> I wish there was a stronger way to prevent this.

This is merely preference of maintainer of this repository to not have a verbose `paket.depedencies` file, we could specify version constraints with nice constraint dsl supporting all the operators one would expect (you can see simple example: http://fsprojects.github.io/Paket/dependencies-file.html) and in any case the full resolution of locked dependencies are readable at all times, in a single place: in `paket.lock` (including transitive dependencies that we don't have to specify in `paket.dependencies`).

As you were explaining writing the qualified name in the *proj files is recommended in certain circumstances, I've open a feature request for paket: https://github.com/fsprojects/Paket/issues/1823 which will eventually add an option to have that behaviour, thanks a lot for your input!

Issue #1426, opened 2016-07-15
Functional wrappers to make VS SDK code easier to cope with


smoothdeveloper:
I'm looking at few usage of SnapshotSpan / SnapshotPoint and it seems:

1. we don't use properly the VS API
1. we could make few F# helpers to keep API more F# idiomatic

### proper usage of VS API

in our existing code:

```fsharp
SnapshotSpan(x.Start, (SnapshotPoint (x.Snapshot,x.End.Position + num)))
```

If I'm right this is identical to:

```fsharp
SnapshotSpan(x.Start, x.End + num)
```

I think we should use the shorter instead, it also probably amounts to less allocations and require less tracking of state (passing correct ITextSnapshot, etc.)

### F# Helpers

````
VS.Snapshot.mkSpan x.Start (x.End + num)
```

Would few of those helpers keep our code more friendly to potential contributors who don't have much knowledge of VS API?

This also makes a first step toward a higher level DSL to deal with those while having 0 overhead with usage of small inline functions.

cloudRoutine:
This is a good time to tackle this since we're reworking a large amount of the codebase and these changes can happen with the migrations.

smoothdeveloper:
#1438 implements this

PR #1438, opened 2016-07-16
Small cleaning of usage of SnapshotSpan/SnapshotPoint

smoothdeveloper:
see #1426
* refactor low hanging fruits related to usage of SnapshotPoint
* introduce VS.Snapshot module with helper functions to build SnapshotPoint and SnapshotSpan

### DIFF
--- a/src/FSharp.Editing.VisualStudio/CodeGeneration/ImplementInterfaceSmartTagger.fs
+++ b/src/FSharp.Editing.VisualStudio/CodeGeneration/ImplementInterfaceSmartTagger.fs
@@ -45,8 +45,8 @@ type ImplementInterface
 
     let queryInterfaceState (point: SnapshotPoint) (project: IProjectProvider) =
         asyncMaybe {
-            let line = point.Snapshot.GetLineNumberFromPosition point.Position
-            let column = point.Position - point.GetContainingLine().Start.Position
+            let line = point.Line
+            let column = point.Column
             let! ast = vsLanguageService.ParseFileInProject(doc.FilePath, project)
             let pos = Pos.fromZ line column
             let! input = ast.ParseTree
--- a/src/FSharp.Editing.VisualStudio/Common/Utils.fs
+++ b/src/FSharp.Editing.VisualStudio/Common/Utils.fs
@@ -75,17 +75,20 @@ let isGenericTypeParameter = isTypeParameter '''
 let isStaticallyResolvedTypeParameter = isTypeParameter '^'
 
 type SnapshotPoint with
+    member inline x.Line = x.Snapshot.GetLineNumberFromPosition x.Position
+    member inline x.Column = x.Position - x.GetContainingLine().Start.Position
+    member inline x.LineText = x.GetContainingLine().GetText()
     member x.InSpan (span: SnapshotSpan) = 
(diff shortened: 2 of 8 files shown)


src/FSharp.Editing.VisualStudio/CodeGeneration/ImplementInterfaceSmartTagger.fs
```diff
@@ -45,8 +45,8 @@ type ImplementInterface
 
     let queryInterfaceState (point: SnapshotPoint) (project: IProjectProvider) =
         asyncMaybe {
-            let line = point.Snapshot.GetLineNumberFromPosition point.Position
-            let column = point.Position - point.GetContainingLine().Start.Position
+            let line = point.Line
+            let column = point.Column
```
vasily-kirichenko: Why the intermediate bindings?

src/FSharp.Editing.VisualStudio/Common/Utils.fs
```diff
@@ -111,12 +114,9 @@ type SnapshotSpan with
     member inline x.EndLineNum  = x.Snapshot.GetLineNumberFromPosition x.End.Position
     member inline x.EndColumn = x.End.Position - x.EndLine.Start.Position
 
-    member x.ModStart (num) = SnapshotSpan(SnapshotPoint (x.Snapshot, x.Start.Position + num), x.End)
-    member x.ModEnd (num) = SnapshotSpan(x.Start, (SnapshotPoint (x.Snapshot,x.End.Position + num)))
-
-    member x.ModBoth m1 m2 =
-        SnapshotSpan(SnapshotPoint (x.Snapshot, x.Start.Position + m1),
-                     SnapshotPoint (x.Snapshot, x.End.Position + m2))
+    member x.ModStart num = VS.Snapshot.mkSpan (x.Start + num) x.End
```
vasily-kirichenko: Maybe make `mkSpan` tupled? It seems the arguments are often enclosed in parens anyway, so making it tupled could make the code clearer.

src/FSharp.Editing.VisualStudio/CodeGeneration/ImplementInterfaceSmartTagger.fs
smoothdeveloper: @vaskir: line and column used several times bellow.

src/FSharp.Editing.VisualStudio/Common/Utils.fs
smoothdeveloper: I'd suggest we use tupled only when there are overloads, would that be fine?

vasily-kirichenko: I just think that
```fsharp
VS.Snapshot.mkSpan (x.Start + m1, x.End + m2)
```
is better than
```fsharp
VS.Snapshot.mkSpan (x.Start + m1) (x.End + m2)
```
because there are less parens.

smoothdeveloper: What about other calls? I even think the later one in your example is more readable and more F# idiomatic.

If you want to use parens: 

`SnapshotSpan(x.Start + m1, x.End + m2)` already does that so I don't see use in the module doing the same.

PR #1443, opened 2016-07-17
Try to fix unit tests on Travis CI

dungpa:


tests/FSharp.Editing.Tests/TestHelpers.fs
```diff
@@ -78,35 +78,41 @@ module Collection =
 module LanguageServiceTestHelper =
     open Microsoft.FSharp.Compiler.SourceCodeServices
 
+    let references = 
+        [ yield typeof<System.Object>.Assembly.Location; // mscorlib
```
smoothdeveloper: Good approach :)

maybe refactor to a `Type list |> getAssemblyLocation`?

build.fsx
```diff
@@ -303,6 +303,11 @@ Target "TravisCI" (fun _ ->
   ]
   |> MSBuildRelease "" "Rebuild"
   |> ignore
+  
+  let additionalFiles =
```
smoothdeveloper: I think a comment on reason (name of tests that would fail?) would be great.

dungpa: I will. Th whole thing is trial and error for now.

smoothdeveloper: Yes sorry, it should be obvious. I can help running / debugging that stuff on mac at some point.

Thanks for giving a hand!

tests/FSharp.Editing.Tests/GoToDefinitionTests.fs
```diff
@@ -51,7 +51,13 @@ let tryGenerateDefinitionFromPos caretPos src =
...
         let! parseTree = parseResults.ParseTree
         let getXmlDocBySignature = 
-            let xmlFile = symbolUse.Symbol.Assembly.FileName |> Option.map (fun fileName -> Path.ChangeExtension(fileName, ".xml"))
+            let xmlFile = 
+                symbolUse.Symbol.Assembly.FileName 
+                |> Option.map (fun fileName -> 
+                    if fileName.EndsWith "mscorlib.dll" then 
+                        Path.Combine(__SOURCE_DIRECTORY__, "../../lib/mscorlib.xml")
+                    else
+                        Path.ChangeExtension(fileName, ".xml"))
```
cloudRoutine: we may need to check for upper case path extensions [like how FSAC does](https://github.com/fsharp/FsAutoComplete/blob/master/src/FsAutoComplete.Core/TipFormatter.fs#L54-L67) due to linux's case sensitive paths.

tests/FSharp.Editing.Tests/GoToDefinitionTests.fs
```diff
@@ -260,7 +260,7 @@ type Console =
     static member Beep : unit -> unit
 """
 
-[<Test>]
+[<Test; Ignore "Need extra xml files on Travis">]
```
smoothdeveloper: @dungpa do we want to do Assert.Inconclusive instead? This would enable to keep the test runing and be either green or ignored.

We could also use Inconclusive only when we are running through Travis so it would still catch regressions on AppVeyor.

tests/FSharp.Editing.Tests/GoToDefinitionTests.fs
dungpa: That's a good idea. Thanks.

tests/FSharp.Editing.Tests/GoToDefinitionTests.fs
dungpa: I think those tests should be in a separate category and should be ignored on Travis.

dungpa:
@smoothdeveloper Could you take a look at these failing tests related to LoadDirective https://travis-ci.org/fsprojects/VisualFSharpPowerTools/builds/145691414#L1301-L1308?

dungpa:
Down to 8 failing tests. We're close.

cloudRoutine:
When I use goto def on F# list in VS I get 
```fsharp
namespace Microsoft.FSharp.Collections

open Microsoft.FSharp.Collections

/// The type of immutable singly-linked lists.
[<DefaultAugmentation(false)>]
[<StructuralEquality>]
[<StructuralComparison>]
[<CompiledName("FSharpList`1")>]
type List<'T> =
    | ( [] )
    | ( :: ) of Head: 'T * Tail: 'T list
    interface System.Collections.IEnumerable
    interface System.Collections.Generic.IEnumerable<'T>
    /// Gets a slice of the list, the elements of the list from the given start index to the given end index.
    /// startIndex: The start index.
    /// endIndex: The end index.
    member GetSlice : startIndex:int option * endIndex:int option -> 'T list
    /// Gets the first element of the list
    member Head : 'T
    /// Gets a value indicating if the list contains no entries
    member IsEmpty : bool
    /// Gets the element of the list at the given position.
    /// index: The index.
    member Item : 'T
    /// Gets the number of items contained in the list
    member Length : int
    /// Gets the tail of the list, which is a list containing all the elements of the list, excluding the first element
    member Tail : 'T list
    /// Returns a list with head as its first element and tail as its subsequent elements
    /// head: A new head value for the list.
    /// tail: The existing list.
    static member Cons : head:'T * tail:'T list -> 'T list
    /// Returns an empty list of a particular type
    static member Empty : 'T list
```
which doesn't match the hard coded expected result. 
I think the issue here is that Appveyor is using F# 3.1 and Travis is using F# 4.0

cloudRoutine:
It seems like this also applies to the `go to metadata from module and module function`, I bet this is the pervasive issue for the remaining failing tests.

VS generated metadata -
```fsharp
/// Basic operations on options.
[<CompilationRepresentation(enum<CompilationRepresentationFlags> (4))>]
module Microsoft.FSharp.Core.Option
open System

/// Returns true if the option is not None.
/// option: The input option.
val isSome : option:'T option -> bool
/// Returns true if the option is None.
/// option: The input option.
val isNone : option:'T option -> bool
/// Gets the value associated with the option.
/// option: The input option.
val get : option:'T option -> 'T
/// count inp evaluates to match inp with None -> 0 | Some _ -> 1.
/// option: The input option.
val count : option:'T option -> int
/// fold f s inp evaluates to match inp with None -> s | Some x -> f s x.
/// folder: A function to update the state data when given a value from an option.
/// state: The initial state.
/// option: The input option.
val fold : folder:('State -> 'T -> 'State) -> state:'State -> option:'T option -> 'State
/// fold f inp s evaluates to match inp with None -> s | Some x -> f x s.
/// folder: A function to update the state data when given a value from an option.
/// option: The input option.
/// state: The initial state.
val foldBack : folder:('T -> 'State -> 'State) -> option:'T option -> state:'State -> 'State
/// exists p inp evaluates to match inp with None -> false | Some x -> p x.
/// predicate: A function that evaluates to a boolean when given a value from the option type.
/// option: The input option.
val exists : predicate:('T -> bool) -> option:'T option -> bool
/// forall p inp evaluates to match inp with None -> true | Some x -> p x.
/// predicate: A function that evaluates to a boolean when given a value from the option type.
/// option: The input option.
val forall : predicate:('T -> bool) -> option:'T option -> bool
/// iter f inp executes match inp with None -> () | Some x -> f x.
/// action: A function to apply to the option value.
/// option: The input option.
val iter : action:('T -> unit) -> option:'T option -> unit
/// map f inp evaluates to match inp with None -> None | Some x -> Some (f x).
/// mapping: A function to apply to the option value.
/// option: The input option.
val map : mapping:('T -> 'U) -> option:'T option -> 'U option
/// bind f inp evaluates to match inp with None -> None | Some x -> f x
/// binder: A function that takes the value of type T from an option and transforms it into an option containing a value of type U.
/// option: The input option.
val bind : binder:('T -> 'U option) -> option:'T option -> 'U option
/// filter f inp evaluates to match inp with None -> None | Some x -> if f x then Some x else None.
/// predicate: A function that evaluates whether the value contained in the option should remain, or be filtered out.
/// option: The input option.
val filter : predicate:('T -> bool) -> option:'T option -> 'T option
/// Convert the option to an array of length 0 or 1.
/// option: The input option.
val toArray : option:'T option -> 'T []
/// Convert the option to a list of length 0 or 1.
/// option: The input option.
val toList : option:'T option -> 'T list
/// Convert the option to a Nullable value.
/// option: The input option.
val toNullable : option:'T option -> Nullable<'T> when 'T : (new : unit -> 'T) and 'T : struct and 'T :> ValueType
/// Convert a Nullable value to an option.
/// value: The input nullable value.
val ofNullable : value:Nullable<'T> -> 'T option when 'T : (new : unit -> 'T) and 'T : struct and 'T :> ValueType
/// Convert a potentially null value to an option.
/// value: The input value.
val ofObj : value:'T -> 'T option when 'T : null
/// Convert an option to a potentially null value.
/// value: The input value.
val toObj : value:'T option -> 'T when 'T : null
```

cloudRoutine:
appveyor is failing due to the metadata strings being generated with platform specific (windows `\r\n`) newlines

cloudRoutine:
@dungpa  you have the right directory for `mscorlib` on mono for the Travis build, but none of the xml files are stored in that dir. They have their own weird approach to documentation and I'm not sure there's an easy way to get the xml files we're looking for.  I've been asking around and trying to find a viable solution, but haven't had any luck. I've haven't checked the monodevelop codebase yet, but the answer may be tucked away in that src.

dungpa:
Both AppVeyor and Travis are green now (The failing test is not related.).

vasily-kirichenko:
it was easy ;)

PR #1451, opened 2016-07-23
fix rename GUI, fix outdated warnings

cloudRoutine:


vasily-kirichenko:
I don't like the way wpf works: everything compiles fine, but crashes at runtime :(

PR #1455, opened 2016-07-23
use nuget packages instead of ...

smoothdeveloper:
... GAC assemblies for few references in TestUtilities.csproj, I have no clue where the remaining ones on nuget.


cloudRoutine:
LGTM

PR #1458, opened 2016-08-01
cleanup project files

cloudRoutine:
- fix root namespaces and project names
- auto update doc files with assembly name changes

### DIFF
--- a/src/FSharp.Editing.VisualStudio.v2015/FSharp.Editing.VisualStudio.v2015.fsproj
+++ b/src/FSharp.Editing.VisualStudio.v2015/FSharp.Editing.VisualStudio.v2015.fsproj
@@ -7,25 +7,22 @@
     <SchemaVersion>2.0</SchemaVersion>
     <ProjectGuid>bb0c6f01-5e57-4575-a498-5de850d9fa6c</ProjectGuid>
     <OutputType>Library</OutputType>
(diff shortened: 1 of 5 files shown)


src/FSharp.Editing.VisualStudio.v2015/FSharp.Editing.VisualStudio.v2015.fsproj
```diff
@@ -7,25 +7,22 @@
...
-    <ReferencePath>
-    </ReferencePath>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <Tailcalls>false</Tailcalls>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>bin\$(Configuration)</OutputPath>
```
smoothdeveloper: those can be moved out of this propertygroup to the previous to remove duplication, same for the release one.

cloudRoutine: I purposefully didn't do that because I'm going to try some experiments with different configurations and I'd end up adding these lines back in.

PR #1472, opened 2016-10-17
Update FCS to v.8 and all the packages as well

vasily-kirichenko:
Fantomas must be updated to use FCS 8 as well, so it should not be merged until https://github.com/dungpa/fantomas/issues/203 is closed and Fantomas version referenced in this PR updated, too.


vasily-kirichenko:
I updated Fantomas, formatting works in tests and reality.

FSharpLint does not work neither in tests nor in IDE.

PR #1480, opened 2016-10-31
Revert "Fixes to make sure that VFPT color scheme updates properly when VS theme is changed"

surban:
Please revert for now, I will look for a proper fix when I have time.


vasily-kirichenko:
Thanks!

PR #1493, opened 2017-01-06
Clarify 'unused identifier' rules in FAQ

piers7:
I didn't understand this when I read this originally, and the penny didn't drop till I read the original issue discussion. I've reworded the relevant section in the FAQ to hopefully make it clearer.


vasily-kirichenko:
Thanks for this, but this project is deprecated and all the work is being moved to Visual F# Tools repository (as readme.md states (in a rather large font :), btw). Please, post requests and bug reports there (VS 2017 only, the earlier versions are not supported actually because the IDE part has been rewritten with Roslyn API).

PR #1501, opened 2019-06-28
Add link to Visual F# repo

milbrandt:
and add hint that the new repo is also for newer Version of Visual Studio, eg VS2019


vasily-kirichenko:
Thank you :)

